PR #191, opened 2022-08-09
Add optional clustering calculations script

### LINKED ISSUE #67: Modify clustering script to store all results in output SCE

As part of the core pipeline, we will want to be able to output a SCE object with clustering assignments included in the `colData` slot of the SCE. Based on what we know about clustering, there is not one size fits all, so we will want to have the capability of including results from multiple types of clustering algorithms (i.e. k-means and graph-based) that have been tested on a range of parameters.  

Here, we will need to create an Rscript that takes as input a normalized SCE with reduced dimensions already calculated. The script will then perform both graph based and k-means clustering on a range of parameters. Additionally, we will also calculate the cluster purity, silhouette width, and check the stability of the clusters, storing all information in the SCE object so that it can be used for plotting when creating an output html report (a separate issue). The output file should be an SCE object with all clustering stats appended. 

The current code that is used in `04-clustering.Rmd` can be converted to an Rscript and we should require the file path to the SCE object and a seed as the input. The output should be the modified SCE object. 

Another thought that we might consider is adding an argument to pick the type of clustering. By default we should perform both clustering methods, but we might want to allow users the ability to discern which type of clustering they would like.

cbethell:
**Issue Addressed**
Closes #67

The purpose of this PR is to start the optional clustering module off with a calculations script, where all of the relevant calculations included in [`04-clustering.Rmd`](https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/main/04-clustering.Rmd) are done here to prepare for the template notebook where the plotting will be done.

This PR adds a `optional-clustering-analyses/clustering-calculations.R` script which takes in the input SCE object, cluster type, and nearest neighbors range/increment to perform the specified graph-based clustering on the range of nearest neighbor values.

Then the cluster validity and stability stats are calculated using the same range of values, then summarized and joined into one table to allow for an output TSV file with a table of these cluster validity/stability stats.

I know that implementing "2-way kmeans" was also mentioned on the associated issue https://github.com/AlexsLemonade/scpca-downstream-analyses/issues/67#issuecomment-1089302268, but I did not include it here to keep things a bit simpler for review since we have been moving along with just the graph-based clustering thus far. Let me know if you think it should be added here however.

- Does this PR seem to accurately address the updated comments on #67?
- Does this output stats table seem suffice?
- Do you have any suggestions for refactoring?

For testing this script, I used the Gawad data, supplying the following values to the param options (for reference):

```
opt$sce <- "../results/scpca_processed_data/SCPCS000216/SCPCL000290_miQC_processed_sce.rds"
opt$seed <- 2021
opt$cluster_type <- "louvain"
opt$nearest_neighbors_range <- c(5:25)
opt$nearest_neighbors_increment <- 5
opt$output_sce_filepath <- "../results/scpca_processed_data/SCPCS000216/SCPCL000290_miQC_processed_sce.rds"
opt$output_stats_filepath <- "../results/scpca_processed_data/SCPCS000216/SCPCL000290_clustering_stats.tsv"
opt$project_root <- ".."
```

**Checklist**


optional-clustering-analysis/clustering-calculations.R
```diff
@@ -0,0 +1,208 @@
...
+  ),
+  optparse::make_option(
+    c("-n", "--nearest_neighbors_range"),
+    type = "integer",
+    default = 5:25,
+    help = "Range with number of nearest neighbors to include during graph 
+    construction. Can be a range of values, e.g. 5:25, or a single value."
+  ),
+  optparse::make_option(
+    c("-i", "--nearest_neighbors_increment"),
```
sjspielman, line 62: `-i` is repeated! You already used it for the input file. I'd remove the short option here and just keep the long.

optional-clustering-analysis/clustering-calculations.R
```diff
@@ -0,0 +1,208 @@
...
+# Read in normalized sce object
+sce <- readr::read_rds(opt$sce)
+
+# Check that the input data contains an SCE with PCA results 
+if(is(sce,"SingleCellExperiment")){
+  if(!"PCA" %in% reducedDimNames(sce)) {
+    stop("PCA results are not found in the provided SingleCellExperiment object.")
+  }
+} else {
+  stop("file specified by --sce (-s) must contain a SingleCellExperiment object.")
```
sjspielman, line 147: This conflicts. You used `-i` for the short opt, but `-s` for the short seed opt. 
```suggestion
  stop("file specified by --sce (-i) must contain a SingleCellExperiment object.")
```

optional-clustering-analysis/clustering-calculations.R
```diff
@@ -0,0 +1,208 @@
...
+validity_stats_df <- create_metadata_stats_df(sce, 
+                                              opt$nearest_neighbors_range, 
+                                              opt$nearest_neighbors_increment, 
+                                              opt$cluster_type)
+
+# Write output file with all cluster validity stats
+readr::write_tsv(validity_stats_df,
+                 file.path(
+                   stats_output_dir,
+                   paste0(opt$library_id, "_clustering_all_stats.tsv")
```
sjspielman, line 177: Sorry if this was already addressed, but should this file maybe be named `_clustering_all_validity_stats.tsv`? The current name implies to me that stability stats are also in this file.

optional-clustering-analysis/clustering-calculations.R
sjspielman:
```suggestion
    c("--nearest_neighbors_increment"),
```

optional-clustering-analysis/clustering-calculations.R
cbethell: Good point! -- will update this 👍

cbethell:
Thank for the re-review @sjspielman! I've implemented those comments you left and re-ran things to ensure the expected output files are produced, see the new `clustering_stats` folder structure below and let me know what you think now!
Screen Shot 2022-08-23 at 3 28 39 PM

sjspielman (approved):

PR #203, opened 2022-08-18
Add overwrite option to optional clustering calculations script

### LINKED ISSUE #193: Add overwrite option to optional clustering script

Upon development of the optional clustering module calculations script (filed in PR #191) and the subsequent template report (#179), it became obvious that re-running the calculations script with a different `params_range` would not overwrite the previous results store in the SCE object, but instead just add to them -- which is not ideal for development and folks who would be trying various `params_range`.

We should likely add an `--overwrite` flag to the `optional-clustering-analysis/clustering-calculations.R` script to allow users to overwrite any existing cluster assignments.

Adding an `--overwrite` param option means, if it is supplied at the command line (or in the Snakemake rule), any existing clustering results will be overwritten. This means that there should also be a check for whether or not there ARE cluster assignments that already exist when we read in the SCE object.

cbethell:
**Issue Addressed**
Closes #193

As mentioned in https://github.com/AlexsLemonade/scpca-downstream-analyses/issues/193#issue-1334863573,  we want to add an `--overwrite` flag to the `clustering-calculations.R` script to allow for users to overwrite any existing cluster assignments (and for ease of development).

This PR adds the `--overwrite` param option to the clustering calculations script, using the `store_true` action.
There is check for whether or not `--overwrite` is NULL, if it is not, the clustering will take place with a note that the existing clustering results are being overwritten, if it _is_ NULL, the script will stop with a message that there are existing clustering results and to use the flag if they should be overwritten.

Does this seem like the best way to tackle #193? Do you have suggestions for refactoring?

Note that this PR is stacked on PR #191

**Checklist**


allyhawkins (commented):
So what I think you actually want to do here is first check if the clustering results (using the provided parameters) already exist or not. Then if they exist you would check for the overwrite option. If they don't exist then you perform clustering regardless of the overwrite option. 

The way that I would probably approach checking that the clustering results are present is through creating the cluster column names given the provided cluster type and param range. Then check to see if those cluster column names are all present in the SCE object. We might want to consider a case for if only some of them are present too and how we would deal with that?

cbethell:
> So what I think you actually want to do here is first check if the clustering results (using the provided parameters) already exist or not. Then if they exist you would check for the overwrite option. If they don't exist then you perform clustering regardless of the overwrite option.
> 
> The way that I would probably approach checking that the clustering results are present is through creating the cluster column names given the provided cluster type and param range. Then check to see if those cluster column names are all present in the SCE object. We might want to consider a case for if only some of them are present too and how we would deal with that?

Good catch @allyhawkins -- I've updated the script in https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/203/commits/7939c8bc8cbd094a6a012e5e388e4a92211cd045 to check if the cluster results already exist like you suggested, and also tried to handle a case for if only some are present by using a for loop and defining the cluster name within the for loop. Let me know what you think or if you have any refactoring suggestions!

optional-clustering-analysis/clustering-calculations.R
```diff
@@ -149,11 +154,50 @@ if(is(sce,"SingleCellExperiment")){
...
+    }
+  } else if (is.null(colData(sce)[[cluster_column_name]])) {
+    sce <- graph_clustering(
+      normalized_sce = sce,
+      params_range = opt$nearest_neighbors_range,
+      step_size = opt$nearest_neighbors_increment,
+      cluster_type = opt$cluster_type
+    )
+  }
+}
```
allyhawkins, line 200: I don't think what you had here will quite work the way you think it will. By having the for loop over the values in the nn range but then inputting the range and increment into the clustering function, you are performing clustering with the same range/ increment multiple times, once for every number in the range. However the actual function isn't changing so you are overwriting clustering results over and over within the for loop. 
Instead I would actually create a vector of the expected column names for the clustering results and then intersect those with the colData column names. If any of them intersect then you can go ahead and check the overwrite option. If not, proceed with clustering as normal. 

In my suggestion I also used `glue::glue` to print out each of the column names that will be overwritten before printing the stop message to use overwrite. 

```suggestion
# Check for existing clustering results
cluster_column_names <- paste(opt$cluster_type, nn_rage, sep = "_")
existing_columns <- intersect(cluster_column_names, colnames(coldata(sce)))
if(length(existing_columns) != 0){
  
  if(!is.null(opt$overwrite)){
    # Perform graph-based clustering
    message("Overwriting clustering results.")
    sce <- graph_clustering(
      normalized_sce = sce,
      params_range = opt$nearest_neighbors_range,
      step_size = opt$nearest_neighbors_increment,
      cluster_type = opt$cluster_type)
  } else {
    message(
      glue::glue("
      Clustering results exist for {cluster_column_name}.")
    )
    stop("Skipping clustering steps. If you want to overwrite the existing results, use the --overwrite flag.")
  }
} else {
  sce <- graph_clustering(
    normalized_sce = sce,
    params_range = opt$nearest_neighbors_range,
    step_size = opt$nearest_neighbors_increment,
    cluster_type = opt$cluster_type)
}
```

allyhawkins (commented):
I think you're on the right track, but I left some suggestions to clean up the for loop you had to make sure that you aren't repeating clustering.

cbethell:
> I don't think what you had here will quite work the way you think it will. By having the for loop over the values in the nn range but then inputting the range and increment into the clustering function, you are performing clustering with the same range/ increment multiple times, once for every number in the range. However the actual function isn't changing so you are overwriting clustering results over and over within the for loop. Instead I would actually create a vector of the expected column names for the clustering results and then intersect those with the colData column names. If any of them intersect then you can go ahead and check the overwrite option. If not, proceed with clustering as normal.
> 
> In my suggestion I also used `glue::glue` to print out each of the column names that will be overwritten before printing the stop message to use overwrite.

Thanks for that fix @allyhawkins -- the logic there definitely initially threw me for a loop (pun intended) but it does make more sense with your changes and looks a lot cleaner! I was able to test that the refactoring changes work successfully so let me know what you think now!

allyhawkins (approved):
LGTM!

PR #202, opened 2022-08-18
Update R to 4.2 & Bioconductor to 3.15

jashapiro:
**Issue Addressed**
closes #183

**What is the purpose of these changes?**

The main goal here is to bring this repo more in sync with `sc-data-integration`, incorporating any updates that have come along with a new release of bioconductor.

**What changes did you make?**

- Update the environment to usr R 4.2. Since some packages are not available in conda, those are now installed by R in the setup script.
- update the `renv.lock` file with new package versions and requirements. There are a few changes in which packages are installed, but those may reflect changes in dependencies. 
- updated docs to reflect new requirements as stored in the `renv.lock` file for non-conda installation. 
 

**Were there any other solutions that you tried?**

Not as now. I was thinking a bit about whether we should maybe retain the old renv.lock file to allow some backward compatibility, but I hadn't quite explored how to do that.

The main reason to do that would be for development on the server, where we currently have R4.1. But since we are recommending the conda environments, _running_ the workflow there should still work.

**Any comments, concerns, or questions important for reviewers**

As previously, if testing on an Apple Silicon Mac, you will want to run 
```
CONDA_SUBDIR=osx-64 snakemake  -c1 --use-conda  --conda-create-envs-only build_renv
```
to build the new environment.

On an Intel Mac, the first run of the workflow should install the new enviroment automatically.

Building the new environment may be slow, but maybe will use a cache if you are lucky.

**Checklist**

Place an `x` in all boxes that you have completed.

- [x] I have run the most recent version of the code
- [x] I have added all necessary documentation (if applicable)

PR #204, opened 2022-08-23
fix extra slash

jashapiro:
This quick hotfix PR is to remove an extra slash that was disrupting a full run of the workflow. this was discovered when the example results directory was not present, but would likely affect other non-example runs as well. https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/202#issuecomment-1224233458

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -53,7 +53,7 @@ rule filter_data:
     input:
         get_input_rds_files
     output:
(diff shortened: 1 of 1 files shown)


cbethell (approved):
This indeed did the trick -- thanks for the fix @jashapiro!

Issue #206, opened 2022-08-25
Add documentation to clustering template notebook


cbethell:
**Please provide some background on the proposed additions or changes.**

Once #179 and the clustering template notebook is in the `optional-clustering-analysis` directory, we will want to had some clear, but general documentation to the notebook -- especially surrounding the plots and how one may interpret the plots.

**What are the changes that you are proposing?**

We will want to add documentation to the clustering template notebook that will help folks make decisions about their results, for example we will want to include the expected ranges for each of the plots, how they are generally interpreted, and likely some links to further documentation that may be beneficial.

**Please describe the proposed solution.**

The idea is that we will add this documentation after adding the main structure of the clustering template notebook (#179), for ease of review (and development).

cbethell:
This was addressed by merged PR #210 👍

PR #207, opened 2022-08-29
Add optional clustering report template

### LINKED ISSUE #179: Add clustering template notebook

Once we have completed #67 and have a script that performs all of the necessary clustering calculations and saves them in an output SCE object, we will want to have clustering template notebook to take in the SCE and plot/display the results.

Add a `clustering/clustering-report-template.Rmd` file that takes in a SCE object containing clustering calculations, and plots the UMAP coordinates for each parameter within the provided range, the cluster validity stats, and the cluster stability stats (similarly to what we do in [`04-clustering.Rmd`](https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/main/04-clustering.Rmd). 

Add this `clustering-report-template.Rmd` file to a modular directory named "clustering", and treat it as a template file which accepts input parameters for processing.

Not that I can think of at the time, but folks are welcome to leave thoughts here!

cbethell:
**Issue Addressed**
Closes #179 

This PR adds the report template for the optional clustering analysis module.

This PR reads in the processed SCE and results files generated using the optional clustering calculations R script and uses these results to generate UMAP, cluster validity, and cluster stability plots.

**Edit:** The template notebook in this PR, more specifically, includes checks as each result file is read in to ensure that the defined `existing_columns` with the clustering results do indeed exist in each of the results files. There is also some logic added to define the figure heights and widths where applicable, as things get get a bit _squished_ if there a large number of parameters, which I found out during development.

This PR also adds a function to the `utils/clustering-functions.R` script named `define_nn_range` which takes the range and increment parameters to define a nearest neighbors range (I also propagated this function over to the `optional-clustering-analysis/clustering-calculations.R` script).

- Does this template seem to suffice issue #179?
- Do we think that facetted plots are ideal for multiple clustering parameters or do we want one report per param? i.e a different report for `louvain_10` vs `louvain_15`?

See a zip of the current example report below:
[SCPCL000290_optional_clustering_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/9461599/SCPCL000290_optional_clustering_report.html.zip)

**Checklist**

- [ ] I have added all necessary documentation (if applicable) -- will be added in a subsequent PR

### DIFF
--- a/optional-clustering-analysis/clustering-calculations.R
+++ b/optional-clustering-analysis/clustering-calculations.R
@@ -12,7 +12,8 @@
 #   --library_id "library01" \
 #   --seed 2021 \
 #   --cluster_type "louvain" \
-#   --nearest_neighbors_range 5:25 \
+#   --nearest_neighbors_lowest 5
+#   --nearest_neighbors_highest 25 \
 #   --nearest_neighbors_increment 5 \
 #   --output_directory "example-results/sample01"
 
@@ -52,11 +53,18 @@ option_list <- list(
     help = "Method used for clustering. Can be either louvain or walktrap.",
   ),
   optparse::make_option(
-    c("-n", "--nearest_neighbors_range"),
+    c("-n", "--nearest_neighbors_lowest"),
     type = "integer",
-    default = 5:25,
-    help = "Range with number of nearest neighbors to include during graph 
-    construction. Can be a range of values, e.g. 5:25, or a single value."
+    default = 5,
+    help = "The lowest number of a range of nearest neighbors values to include 
+    during graph construction. Default is 5."
+  ),
+  optparse::make_option(
+    c("--nearest_neighbors_highest"),
+    type = "integer",
+    default = 25,
+    help = "The highest number of a range of nearest neighbors values to include 
+    during graph construction. Default is 25."
   ),
   optparse::make_option(
     c("--nearest_neighbors_increment"),
@@ -155,7 +163,9 @@ if(is(sce,"SingleCellExperiment")){
 #### Perform clustering --------------------------------------------------------
 
 # Define nearest neighbors range of values
-nn_range <- define_nn_range(opt$nearest_neighbors_range, opt$nearest_neighbors_increment)
+nn_range <- define_nn_range(opt$nearest_neighbors_lowest,
+                            opt$nearest_neighbors_highest,
+                            opt$nearest_neighbors_increment)
 
 # Check for existing clustering results
 cluster_column_names <- paste(opt$cluster_type, nn_range, sep = "_")
@@ -168,7 +178,8 @@ if(length(existing_columns) != 0){
     message("Overwriting clustering results.")
     sce <- graph_clustering(
       normalized_sce = sce,
-      params_range = opt$nearest_neighbors_range,
+      nearest_neighbors_lowest = as.integer(opt$nearest_neighbors_lowest),
+      nearest_neighbors_highest = as.integer(opt$nearest_neighbors_highest),
       step_size = opt$nearest_neighbors_increment,
       cluster_type = opt$cluster_type)
   } else {
@@ -181,7 +192,8 @@ if(length(existing_columns) != 0){
 } else {
   sce <- graph_clustering(
     normalized_sce = sce,
-    params_range = opt$nearest_neighbors_range,
+    nearest_neighbors_lowest = as.integer(opt$nearest_neighbors_lowest),
+    nearest_neighbors_highest = as.integer(opt$nearest_neighbors_highest),
     step_size = opt$nearest_neighbors_increment,
     cluster_type = opt$cluster_type)
 }
@@ -197,8 +209,7 @@ readr::write_rds(sce, file.path(
 # Check the cluster validity stats for each of the clusters in the SCE object
 # and return stats in a data frame
 validity_stats_df <- create_metadata_stats_df(sce, 
-                                              opt$nearest_neighbors_range, 
-                                              opt$nearest_neighbors_increment, 
+                                              nn_range, 
                                               opt$cluster_type)
 
 # Write output file with all cluster validity stats
@@ -225,8 +236,7 @@ readr::write_tsv(summary_validity_stats_df,
 summary_stability_stats_df <-
   get_cluster_stability_summary(
     sce,
-    opt$nearest_neighbors_range,
-    opt$nearest_neighbors_increment,
+    nn_range,
     opt$cluster_type
   ) %>%
   dplyr::mutate(param_value = as.numeric(param_value))
--- a/optional-clustering-analysis/clustering-calculations.R
+++ b/optional-clustering-analysis/clustering-calculations.R
@@ -12,8 +12,8 @@
 #   --library_id "library01" \
 #   --seed 2021 \
 #   --cluster_type "louvain" \
-#   --nearest_neighbors_lowest 5
-#   --nearest_neighbors_highest 25 \
+#   --nearest_neighbors_min 5
+#   --nearest_neighbors_max 25 \
 #   --nearest_neighbors_increment 5 \
 #   --output_directory "example-results/sample01"
 
@@ -53,17 +53,17 @@ option_list <- list(
     help = "Method used for clustering. Can be either louvain or walktrap.",
   ),
   optparse::make_option(
-    c("-n", "--nearest_neighbors_lowest"),
+    c("-n", "--nearest_neighbors_min"),
     type = "integer",
     default = 5,
-    help = "The lowest number of a range of nearest neighbors values to include 
+    help = "The minimum number of a range of nearest neighbors values to include 
     during graph construction. Default is 5."
   ),
   optparse::make_option(
-    c("--nearest_neighbors_highest"),
+    c("--nearest_neighbors_max"),
     type = "integer",
     default = 25,
-    help = "The highest number of a range of nearest neighbors values to include 
+    help = "The maximum number of a range of nearest neighbors values to include 
     during graph construction. Default is 25."
   ),
   optparse::make_option(
@@ -163,8 +163,8 @@ if(is(sce,"SingleCellExperiment")){
 #### Perform clustering --------------------------------------------------------
 
 # Define nearest neighbors range of values
-nn_range <- define_nn_range(opt$nearest_neighbors_lowest,
-                            opt$nearest_neighbors_highest,
+nn_range <- define_nn_range(opt$nearest_neighbors_min,
+                            opt$nearest_neighbors_max,
                             opt$nearest_neighbors_increment)
 
 # Check for existing clustering results
@@ -178,8 +178,8 @@ if(length(existing_columns) != 0){
     message("Overwriting clustering results.")
     sce <- graph_clustering(
       normalized_sce = sce,
-      nearest_neighbors_lowest = as.integer(opt$nearest_neighbors_lowest),
-      nearest_neighbors_highest = as.integer(opt$nearest_neighbors_highest),
+      nearest_neighbors_min = as.integer(opt$nearest_neighbors_min),
+      nearest_neighbors_max = as.integer(opt$nearest_neighbors_max),
       step_size = opt$nearest_neighbors_increment,
       cluster_type = opt$cluster_type)
   } else {
@@ -192,8 +192,8 @@ if(length(existing_columns) != 0){
 } else {
   sce <- graph_clustering(
     normalized_sce = sce,
-    nearest_neighbors_lowest = as.integer(opt$nearest_neighbors_lowest),
-    nearest_neighbors_highest = as.integer(opt$nearest_neighbors_highest),
+    nearest_neighbors_min = as.integer(opt$nearest_neighbors_min),
+    nearest_neighbors_max = as.integer(opt$nearest_neighbors_max),
     step_size = opt$nearest_neighbors_increment,
     cluster_type = opt$cluster_type)
 }
--- a/optional-clustering-analysis/clustering-calculations.R
+++ b/optional-clustering-analysis/clustering-calculations.R
@@ -154,20 +154,8 @@ if(is(sce,"SingleCellExperiment")){
 
 #### Perform clustering --------------------------------------------------------
 
-# If opt$nearest_neighbors_range is a single value, set nearest_neighbors_increment to NULL
-if(length(opt$nearest_neighbors_range) == 1) {
-  opt$nearest_neighbors_increment <- 1
-}
-# If a nearest neighbors increment exists, then create a sequence for clustering 
-if(!(opt$nearest_neighbors_increment == 1)){
-  nn_range <- seq(min(opt$nearest_neighbors_range), 
-                  max(opt$nearest_neighbors_range), 
-                  opt$nearest_neighbors_increment) 
-  # If no nearest neighbors increment has been input then the provided range is 
-  # directly used for clustering 
-} else {
-  nn_range <- opt$nearest_neighbors_range
-}
+# Define nearest neighbors range of values
+nn_range <- define_nn_range(opt$nearest_neighbors_range, opt$nearest_neighbors_increment)
 
 # Check for existing clustering results
 cluster_column_names <- paste(opt$cluster_type, nn_range, sep = "_")
--- a/optional-clustering-analysis/clustering-report-template.Rmd
+++ b/optional-clustering-analysis/clustering-report-template.Rmd
@@ -1,13 +1,12 @@
 ---
 params:
   library: Library
-  processed_sce: "../results/scpca_processed_data/SCPCS000216/SCPCL000290_miQC_processed_sce.rds"
+  processed_sce: "../results/scpca_processed_data/SCPCS000216/SCPCL000290_processed_sce_clustering.rds"
   stats_dir: "../results/scpca_processed_data/SCPCS000216/clustering_stats"
   cluster_type: "louvain"
   nearest_neighbors_range: 5:25
   nearest_neighbors_increment: 5
   project_root: NULL
-  mito_file: "reference-files/Homo_sapiens.GRCh38.104.mitogenes.txt"
   date: !r Sys.Date()
 
 title: "`r glue::glue('Clustering analysis report for {params$library}')`"
@@ -69,10 +68,10 @@ processed_sce <- read_rds(file.path(params$processed_sce))
 
 ## Plots
 
-## Clustering results for `r params$library`
+## UMAP plot results for `r params$library`
 
 Below is a UMAP plot where each cell has been colored based on the cell's cluster assignment. 
-Here clustering was performed using `r params$cluster_type` clustering with a nearest neighbors value of `r params$nearest_neighbors`.
+Here clustering was performed using `r params$cluster_type` clustering.
 
 ```{r}
 # If param$nearest_neighbors_range is a single value, set nearest_neighbors_increment to NULL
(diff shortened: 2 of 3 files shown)


optional-clustering-analysis/clustering-report-template.Rmd
```diff
@@ -0,0 +1,174 @@
+---
+params:
+  library: Library
+  processed_sce: "../results/scpca_processed_data/SCPCS000216/SCPCL000290_miQC_processed_sce.rds"
```
allyhawkins, line 4:
```suggestion
  processed_sce: "../results/scpca_processed_data/SCPCS000216/SCPCL000290_miQC_processed_sce_clustering.rds"
```
I think you actually want the input to this to be the output sce object from the clustering script. https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/2faee9afe5d8ac290ae3546653519b5c2fd735ca/optional-clustering-analysis/clustering-calculations.R#L202-L205

optional-clustering-analysis/clustering-report-template.Rmd
```diff
@@ -0,0 +1,174 @@
...
+  stats_dir: "../results/scpca_processed_data/SCPCS000216/clustering_stats"
+  cluster_type: "louvain"
+  nearest_neighbors_range: 5:25
+  nearest_neighbors_increment: 5
+  project_root: NULL
+  mito_file: "reference-files/Homo_sapiens.GRCh38.104.mitogenes.txt"
```
allyhawkins, line 10:
```suggestion
```
You can remove this. You shouldn't need the mito file for this notebook.

optional-clustering-analysis/clustering-report-template.Rmd
```diff
@@ -0,0 +1,174 @@
...
+                  params$nearest_neighbors_increment) 
+  # If no nearest neighbors increment has been input then the provided range is 
+  # directly used for clustering 
+} else {
+  nn_range <- params$nearest_neighbors_range
+}
+
+# Grab column name with clustering results
+cluster_column_names <- paste(params$cluster_type, nn_range, sep = "_")
+existing_columns <- intersect(cluster_column_names, colnames(colData(processed_sce)))
```
allyhawkins, line 95: I would move this all to a separate chunk above the UMAP section and in that section you want to also perform a check that `length(existing_columns)` is not 0. Otherwise clustering hasn't been performed using the given parameters and the report can't be created. 

Additionally I think it might be time to break out the creation of the `nn_range` and the checks around that into its own function. The function would take as input the `nearest_neighbors_range` and the `nearest_neighbors_increment`, do the steps in line 79-92 and then return the `nn_range`.

optional-clustering-analysis/clustering-report-template.Rmd
```diff
@@ -0,0 +1,174 @@
...
+```
+
+## Plot cluster validity stats
+
+```{r fig.height = 9.5}
+# Grab the cluster validity stats data frame
+all_validity_stats_df <- readr::read_tsv(file.path(
+  params$stats_dir,
+  paste0(params$library, "_clustering_all_validity_stats.tsv")
+))
```
allyhawkins, line 119: After reading this in, I would add a check that makes sure that data corresponding to the `existing_clusters` is present and if it's not there I would throw an error. I also would probably subset the dataframe to those `existing_clusters` cluster names. This should apply to all of the dataframes.

allyhawkins (commented):
The general structure of this is there, I just had some overarching comments to incorporate checks and make sure we are checking that the clustering data is present in the input data.

cbethell:
Thanks for that feedback @allyhawkins! I've added a `define_nn_range` function and also those checks as you suggested and got this ready for an actual review now -- let me know what you think!

Here's the zip of the current example report for reference:
[SCPCL000290_optional_clustering_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/9461618/SCPCL000290_optional_clustering_report.html.zip)

optional-clustering-analysis/clustering-report-template.Rmd
```diff
@@ -0,0 +1,211 @@
...
+
+# Check that the clustering results exist
+if(!(any(existing_columns %in% all_validity_stats_df$cluster_names_column))){
+  stop("The clustering results cannot be found in the all validity stats file.")
+} else {
+  all_validity_stats_df <- all_validity_stats_df %>%
+    dplyr::filter(cluster_names_column %in% existing_columns)
+}
```
allyhawkins, line 125:
```suggestion
stats_column_check <- function(existing_columns, df_column){
  missing_columns <- setdiff(existing_columns, df_column)
  if(length(missing_columns) != 0){
    missing_columns_print <- paste(missing_columns, collapse = ", ")
    stop(glue::glue("The clustering statistics cannot be found for {missing_columns_print}")
    }
  }
stats_column_check(existing_columns, all_validity_stats_df$cluster_names_column)

all_validity_stats_df <- all_validity_stats_df %>%
    dplyr::filter(cluster_names_column %in% existing_columns)
```

So the first thing I would do is create a small subfunction that does the checking for the columns in each of the columns of the dataframes for you so you don't have to repeat the check and error printing everytime. I wrote an example here but the spacing might be a little off so make sure you double check that. 

I also would move these checks and reading in each of the files to the beginning of the notebook after the check for the existing columns in the SCE object. If there are no stats we want to know before we make any UMAPs. 

The last thing I would do is make sure we are checking that all of the clustering results for the whole range of parameters (e.g. all `existing_columns`) have associated statistics. Using `any` will just check that at least one of them is present.

optional-clustering-analysis/clustering-calculations.R
```diff
@@ -154,20 +154,8 @@ if(is(sce,"SingleCellExperiment")){
...
-  nn_range <- seq(min(opt$nearest_neighbors_range), 
-                  max(opt$nearest_neighbors_range), 
-                  opt$nearest_neighbors_increment) 
-  # If no nearest neighbors increment has been input then the provided range is 
-  # directly used for clustering 
-} else {
-  nn_range <- opt$nearest_neighbors_range
-}
+# Define nearest neighbors range of values
+nn_range <- define_nn_range(opt$nearest_neighbors_range, opt$nearest_neighbors_increment)
```
allyhawkins, line 158: Can you also use this function in the `04-clustering.R` script?

allyhawkins (commented):
This is looking pretty good! Just one more refactoring suggestions and then some additional plotting thoughts.

optional-clustering-analysis/clustering-calculations.R
cbethell: This was actually not done in the `04-clustering.R` script but instead in the graph_clustering function in the `utils/clustering-functions.R` script so I implemented the change there.

cbethell:
I've implemented those additional refactoring suggestions @allyhawkins and played around with `point_size` and `point_alpha` a bit -- let me know what you think of the latest changes! 

See the updated html for reference:
[SCPCL000290_optional_clustering_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/9469615/SCPCL000290_optional_clustering_report.html.zip)

I am also going to ping @sjspielman for a review here!

optional-clustering-analysis/clustering-report-template.Rmd
```diff
@@ -0,0 +1,228 @@
+---
+params:
+  library: Library
+  processed_sce: "../results/scpca_processed_data/SCPCS000216/SCPCL000290_processed_sce_clustering.rds"
+  stats_dir: "../results/scpca_processed_data/SCPCS000216/clustering_stats"
```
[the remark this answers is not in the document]
sjspielman: There are further issues here - I went to rerun the Snakefile in order to generate input result files for this notebook. This failed because of how the argument `nearest_neighbors_increment` is being used in `define_nn()`. The `if` statement: `if !(nearest_neighbors_increment == 1)` cannot actually be evaluated if this argument is `NULL`. I've made a suggestion at that line that should do the trick to revive the Snakefile.

optional-clustering-analysis/clustering-report-template.Rmd
```diff
@@ -0,0 +1,228 @@
...
+  html_document:
+    toc: true
+    toc_depth: 2
+    toc_float: true
+    number_sections: true
+    code_folding: hide
+---
+
+This notebook will look at additional clustering parameters on the expression of
+the genes in a single sample of the dataset of interest, from an unbiased approach.
```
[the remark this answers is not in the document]
allyhawkins: Just want to note that we have a separate issue for adding documentation and explanation into the notebook #206 since we want to expand on this a lot more in general which is going to take another round of review.

sjspielman:
@cbethell Ok final comment from first round of review! The notebook won't render for me, and I realized why - it is knitting from the `optional-clustering-analysis` directory and paths _for setting up `renv`_ fail. For previous analysis templates this wasn't a problem because they knit from the project root itself. 

To fix this, you'll need to have _TWO_ setup chunks, something like this:


Chunk 1:
```
knitr::opts_chunk$set(
  message = FALSE,
  warning = FALSE
)

# If project root is not provided use here::here() 
if (is.null(params$project_root)) {
  project_root <- here::here()
} else {
  project_root <- params$project_root
} 

# Source in set up function
source(file.path(project_root, "utils", "setup-functions.R"))
source(file.path(project_root, "utils", "clustering-functions.R"))

# Tell knitr to knit from the project root to be able to load `renv` properly
# This path goes into effect for SUBSEQUENT chunks, hence the need for 2 chunks
knitr::opts_knit$set(root.dir = project_root)

# Set theme for plotting
library(ggplot2)
theme_set(theme_bw() +  theme(text = element_text(size = 12), legend.text = element_text(size=12))) # note this was a suggestion of mine
```

Chunk 2:
```{r more_setup}
# Load project
setup_renv(project_filepath = project_root)
```

sjspielman:
@cbethell Sorry I lied, one more intial review comment! The font size of 12 seems _really_ small - it's hard for me to see labels on the final plots. I recommend tweaking this to a higher value.

optional-clustering-analysis/clustering-report-template.Rmd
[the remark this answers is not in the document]
cbethell: Linking the PR here for reference: https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/210

cbethell:
Thanks for the testing and feedback @sjspielman! I was able to successfully adapt things to take in a single value for a lower param and a higher param and use those to form a nn range as discussed. I also changed over to using the `example-results` files and was able to knit things via RStudio, but it seems that I am still getting that weird behavior now when I go back to running things via command line! See the errors below:


Screen Shot 2022-09-01 at 3 44 45 PM

and when I comment out that `setup_renv()` line: 

Screen Shot 2022-09-01 at 3 44 06 PM

See also an example of the knitted version I was able to get from RStudio with all other changes successfully implemented:

[clustering-report-template.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/9472917/clustering-report-template.html.zip)

cbethell:
**Update:** I was successfully able to run things with the recent renv updates after changing the location of `library(dplyr)` to be before the `setup_renv()` function (https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/207/commits/9786c7231dcdae564465a5a10c828b74a8a20af6) due to the following error:
Screen Shot 2022-09-02 at 8 56 37 AM


I used the following command via command line to successfully run things (where the output file is the only filepath that needs `../` and project root is `NULL`:

```
Rscript -e "rmarkdown::render('clustering-report-template.Rmd', clean = TRUE,
              output_file = '../example-results/sample01/library01_optional_clustering_report.html',
              params = list(library = 'library01',
                            processed_sce = 'example-results/sample01/library01_processed_sce_clustering.rds', stats_dir = 'example-results/sample01/clustering_stats', cluster_type = 'louvain', nearest_neighbors_lowest = 5, nearest_neighbors_highest = 25, nearest_neighbors_increment = 5), 
              envir = new.env())"
```

See the zip of the updated html below:
[library01_optional_clustering_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/9478052/library01_optional_clustering_report.html.zip)

optional-clustering-analysis/clustering-report-template.Rmd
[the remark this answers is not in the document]
cbethell: Filed an issue to address that README update - https://github.com/AlexsLemonade/scpca-downstream-analyses/issues/211

sjspielman:
@cbethell Yes, we definitely want to load libraries _after_ running `setup_renv()`. I should have placed library loads in the "second chunk" in my previous suggestion, sorry about that!!

utils/clustering-functions.R
```diff
@@ -62,8 +62,42 @@ kmeans_clustering <- function(normalized_sce,
...
+  # If no nearest neighbors increment has been input then the provided range is 
+  # directly used for clustering 
+  if (is.null(nearest_neighbors_increment)) {
+    nn_range <- nearest_neighbors_range
+  } else {
+    # If a nearest neighbors increment exists, then create a sequence for clustering 
+    nn_range <- seq(min(nearest_neighbors_range), 
+                    max(nearest_neighbors_range), 
+                    nearest_neighbors_increment) 
+  }
```
allyhawkins, line 93:
```suggestion
  if (is.null(nearest_neighbors_increment)) {
    nearest_neighbors_increment <-1
  } 
  nn_range <- seq(min(nearest_neighbors_range), 
                  max(nearest_neighbors_range), 
                  nearest_neighbors_increment) 

```
So I think here you just need to assign the increment to something and then create the range regardless. I'm not sure what you had will work if the increment was NULL.

optional-clustering-analysis/clustering-report-template.Rmd
```diff
@@ -0,0 +1,236 @@
...
+cowplot::plot_grid(plotlist = plot_list, ncol = num_col)
+```
+
+## Plot cluster validity stats
+
+#### Purity plots
+
+```{r}
+# Plot individual cluster purity stats
+purity_plots <- plot_cluster_purity(all_validity_stats_df, num_col)
```
allyhawkins, line 173: Something about these plots looks off to me. The legend here is really big and almost the center of attention while there are two squished columns with data. I almost think for this one we need three columns...

allyhawkins (commented):
The actual notebook looks fairly good to me, just a small error still in the definition of the nearest neighbors range function that needs to be fixed and the purity plots need some adjusting I think. 

Also regarding the libraries and environment, since we will be running this in a snakefile we can build a conda environment set up that we did with the core workflow so that all of the dependencies are loaded together. But that should come after we build the snakefile as one of the last steps.

utils/clustering-functions.R
cbethell: Ah yes, I think this is it -- this makes sense (will still test for the various possible cases here) 👍

cbethell:
Okay I've moved all of the libraries to be loaded after `setup_renv()`, fixed the plots to be a bit more square in size and also fixed the legend of the purity plots, and replaced/propagated the change from `lowest` and `highest` to `min` and `max`.

Here's the successful output now!
[library01_optional_clustering_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/9479536/library01_optional_clustering_report.html.zip)

utils/clustering-functions.R
```diff
@@ -62,8 +62,42 @@ kmeans_clustering <- function(normalized_sce,
...
+  #                          values to include when calculating/plotting the 
+  #                          clustering results.
+  #   nearest_neighbors_max: maximum number of a range of nearest neighbors 
+  #                          values to include when calculating/plotting the 
+  #                          clustering results.
+  #   nearest_neighbors_increment: increment to use when implementing the range 
+  #                                number of nearest neighbors for cluster stats.
+  
+  # Define range using the min and max values
+  nearest_neighbors_range <- c(nearest_neighbors_min:nearest_neighbors_max)
```
allyhawkins, line 82: I don't think this line is actually doing anything? You're creating the sequence below anyways so you don't need this anymore.

allyhawkins (approved):
One more small change but no need for me to take another look 👍

utils/clustering-functions.R
sjspielman: Agreed; delete this and change below as I suggest!

utils/clustering-functions.R
```diff
@@ -62,8 +62,42 @@ kmeans_clustering <- function(normalized_sce,
...
+  # If no nearest neighbors increment has been input then the provided range is 
+  # directly used for clustering and nearest neighbors increment is set to 1
+  if (is.null(nearest_neighbors_increment)) {
+    nearest_neighbors_increment <- 1
+  }
+  nn_range <- seq(
+    min(nearest_neighbors_range),
+    max(nearest_neighbors_range),
+    nearest_neighbors_increment
+  )
```
sjspielman, line 93:
```suggestion
  nn_range <- seq(
    nearest_neighbors_min,
    nearest_neighbors_max,
    nearest_neighbors_increment
  ) 
```

sjspielman (commented):
I have a couple overall comments about the plots - 

- The font size in summary plots (avg purity and avg width) as well as cluster stability is good, but the _point size_ in the cluster stability plot is little small. Points are very hard to see. Can you increase the size a bit?
- The font size (all around, axes and strip). for UMAPs and full plots for purity and silhouette still seem too small to me. Can you bump these a bit? The point sizes are a bit small too, in particular for UMAP and purity plots where points are often lighter colors. I really have to squint to see them! Tiny bumps here would be good too.

While you're tweaking sizes, it may or may not make more sense to add to the overall theme vs to individual plots, so this is something to play around with a little!

cbethell:
> I have a couple overall comments about the plots -
> 
> * The font size in summary plots (avg purity and avg width) as well as cluster stability is good, but the _point size_ in the cluster stability plot is little small. Points are very hard to see. Can you increase the size a bit?
> * The font size (all around, axes and strip). for UMAPs and full plots for purity and silhouette still seem too small to me. Can you bump these a bit? The point sizes are a bit small too, in particular for UMAP and purity plots where points are often lighter colors. I really have to squint to see them! Tiny bumps here would be good too.
> 
> While you're tweaking sizes, it may or may not make more sense to add to the overall theme vs to individual plots, so this is something to play around with a little!

Okay I added a `point_size` argument to the plotting functions (I couldn't adjust the point size with the theme function as we do with the font size). I also adjusted the overall font size for the first couple of plots, noting that I had to handle the UMAP plots separate from the others as the function there seemed to be overriding what was set at the top of the notebook - I think this should be fine however because we do want the point size and font size there to be a bit smaller than the rest (per this https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/207#discussion_r959756760 regarding the point size, and also because the legends become cut off a bit if we go any larger with the font size). 

Let me know what you think now @sjspielman!

[library01_optional_clustering_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/9499402/library01_optional_clustering_report.html.zip)

sjspielman:
@cbethell The sizes look great all around, thanks! One small thing I think will keep this more robust moving forward is whether we actually want a default for the `point_size` argument in these functions. I'm leaning heavily towards yes, since this argument might be easy to forget. So, maybe let's pop in that 0.7 as a default?

cbethell:
> @cbethell The sizes look great all around, thanks! One small thing I think will keep this more robust moving forward is whether we actually want a default for the `point_size` argument in these functions. I'm leaning heavily towards yes, since this argument might be easy to forget. So, maybe let's pop in that 0.7 as a default?

You're right, I added a default of 0.7 to the plotting functions -- but also thought that this maybe should be one of the params for the template report notebook as well for users to also have the option to tweak -- let me know what you think of this as well!

sjspielman (approved):
Looks great, thanks for the changes! I confirmed this knits both in RStudio and command line.

PR #214, opened 2022-09-08
Initiate Snakefile for optional clustering module - calculations rule

### LINKED ISSUE #180: Initiate Snakefile for clustering module - clustering calculations rule

Once we have completed #67 and #179 within a clustering subdirectory, we will want to add a Snakefile specific to the clustering module. The input to this Snakefile should be SCE with normalized data (ideally the SCE object that has been processed using the core workflow). The first rule of the Snakefile would be to run the clustering calculations script (from #67) on the normalized SCE to calculate and store the clustering results in the SCE.

We will likely want to add a configuration file in addition to the Snakefile, for the purpose of specifying the input, output directory and any other relevant parameters.

**Edit:** This issue should focus on adding the Snakefile and getting the first rule successfully implemented. The first rule here, as mentioned above, should be focused on running the [`clustering-calculations.R`](https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/development/optional-clustering-analysis/clustering-calculations.R) script whose output will be the input to the second rule (to be described in a subsequent issue).

cbethell:
**Issue Addressed**
Closes #180

The purpose of this PR is to add a Snakefile to run the optional clustering module, along with the dependent files including a `config.yaml` and a project metadata file (as we did for the core analysis workflow).

This PR adds the following files to the optional-clustering-analysis module:
- `Snakefile` (to run the clustering-calculations.R script as the first rule)
- `config.yaml` (to set up the parameters required to run the Snakefile)

A `project-metadata/example-library-metadata-clustering.tsv` file has also been added to provide the necessary sample_id, library_id, and filepath to the input processed SCE RDS file.

This PR is a draft because I have not been able to successfully run the Snakefile in this PR due to the following error when I attempt to run using `CONDA_SUBDIR=osx-64 snakemake --use-conda --conda-create-envs-only -c1 build_renv` as well as `snakemake --cores 2 --use-conda` from the optional-clustering-analysis directory:
Screen Shot 2022-09-08 at 2 37 44 PM

Ideally we would use the same conda environment we were able to set up for the core workflow, and interestingly it looks like Snakemake was able to find `envs/scpca-renv.yaml` file, however there seems to be some issue with finding the `.Rprofile` file? Does this seem like the right set up for our intended use of the Snakefile in this PR?

Please let me know if you have any suggestions/thoughts on the error above! cc: @jashapiro @allyhawkins 

**Checklist**


jashapiro:
I suspect that the problem is that the code in the `.Rprofile` file expects to be run from the location of that file (or the root of this repo), so it is looking for an `renv.lock` file that does not exist in this subdirectory. 

There are a few ways we could handle this, but the simplest may be to simply move the clustering module snakefile up a level, renaming it `cluster.snakemake` or something like that, and then invoking it with 

```
snakemake --snakefile cluster.snakemake -c2 --use-conda
```

As you saw, since the environment is the same, you should not need the `build_renv` rule: you can just use the one in `Snakefile` if/when you need to update the environments.

The only other thought is whether you would want to just combine the config files for the two scripts, or if you would want to have a separate one for the clustering workflow. I'd probably put them in the same file unless there are conflicting settings, just for simplicity.

allyhawkins:
I was just looking into this too and I think we want to keep the building of the environment together in the one snakefile and reference the existing environment in the rules that we need it for as Josh said. I was also reading more about [modules](https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html#modules) to see if that's something we would want to explore more, but I think it still would be tricky to manage that unfortunately, so best to keep things separate. 

I would also agree about combining the config files. 

In doing this, we may want to clean up the root directory and move all the scripts for the core workflow into their own folder, if we are going to be putting multiple snakefiles in the root.

cbethell:
Thanks @jashapiro and @allyhawkins!! Adding `cluster.snakefile` to the root directory and combining the config files seemed to work successfully! My last commit included a minor change to the `04-clustering.R` script as well based on some upstream changes in `development`, as I needed this to successfully test and run both snakefiles.

I was able to successfully run the core `Snakefile` using `snakemake -c2 --use-conda` and the `cluster.snakefile` using `snakemake --snakefile cluster.snakefile -c2 --use-conda`.

I also added the core analysis scripts to a `core-analysis` folder, but left the `core-analysis-report-template.Rmd` file in the root directory as I got the following error otherwise:
Screen Shot 2022-09-09 at 9 41 59 AM

config.yaml
```diff
@@ -7,6 +7,7 @@
 results_dir: "example-results/"
 mito_file: "reference-files/Homo_sapiens.GRCh38.104.mitogenes.txt"
 project_metadata: "project-metadata/example-library-metadata.tsv"
+clustering_project_metadata: "project-metadata/example-library-metadata-clustering.tsv"
```
allyhawkins, line 10: One thought I had was rather than have a brand new metadata file for this where we just have the path changed to the output from the core workflow, can we modify the snakefile you have now to take the same metadata file we use for the core workflow and then create the expected file names based on the pattern that we create in the core workflow? Because the core workflow is a requirement I think it's okay to set up that same expected pattern. Here you would just take the library ID and sample ID from the metadata file and then build the full path using `os.path.join(config["results_dir"], {sample_id}/{library_id}_{filtering_method}_processed_sce.rds`. This would replace the input function that you have now in the first rule.

allyhawkins (commented):
I think before I go too far in my review we should simplify the metadata and use the same metadata file for both workflows rather than have users have to create an entirely new one with overlapping information. I left a comment on how I would try and approach doing that. 

As for the error you are seeing, when you moved the Rmd file into a new folder it's now rendering the notebook from that new location and everything is relative to the path of where the notebook lives which would be `{workflow.basedir}/core-analysis`. So it's trying to write the output file to a new directory relative to that folder. So what you actually want to do is change that rule to specify the output directory, `rmarkdown::render(output_dir = '{config[results_dir]}'`.

jashapiro (commented):
This looks good, but I suggested a few changes to let snakemake do the things that snakemake is good at, namely finding the files it needs.

cbethell:
Thank you both for the reviews! Unfortunately I encountered a couple errors when attempting to implement some of the suggestions:

1. @jashapiro, when I incorporated your suggestion to just make the whole `clustering_stats` directory the output (https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/214#discussion_r968801590 and https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/214#discussion_r968806758), I encountered the following error:
Screen Shot 2022-09-13 at 11 03 30 AM

Looks like snakemake isn't happy that we are not using `library_id` when defining the output directory?

2. @allyhawkins, when I incorporated your suggestion to use `os.path.join(config["results_dir"], {sample_id}/{library_id}_{filtering_method}_processed_sce.rds` as the input to the rule in this PR, I got the following error:

Screen Shot 2022-09-13 at 11 19 11 AM

Another wildcard error -- we haven't used `{filtering_method}` in this snakefile as yet when defining the target outputs so looks like snakemake doesn't know what to do with filtering_method. Any thoughts here? Maybe we can add filtering_method to the output filenames (still looking for a snakemake workaround)?

3. I was also able to put the `core-analysis-report-template.Rmd` file in the `core-analysis` subdirectory and run things only after adjusting the `output_dir` in the render function, and adjusting the file paths of the input and output for the rule so that `wildcards.sample_id` would be present for saving the output file in 
`output_dir = '{config[results_dir]}/{wildcards.sample_id}'`

I have confirmed that things run successfully with the updates committed in the most recent commits.

jashapiro:
> Looks like snakemake isn't happy that we are not using `library_id` when defining the output directory?

Yeah, that seems to be the issue. We could make a subdirectory of `clustering_stats` for each library, but it is probably not worth it.

jashapiro:
> 3\. I was also able to put the `core-analysis-report-template.Rmd` file in the `core-analysis` subdirectory and run things only after adjusting the `output_dir` in the render function, and adjusting the file paths of the input and output for the rule so that `wildcards.sample_id` would be present for saving the output file in
>     `output_dir = '{config[results_dir]}/{wildcards.sample_id}'`

You will probably want to use `{wildcards.base_dir}/{wildcards.sample_id}` rather than referring to the config setting in the rule itself.

Snakefile
```diff
@@ -122,16 +122,17 @@ rule clustering:
 rule generate_report:
     input:
         pre_processed_sce = get_input_rds_files,
-        processed_sce =  "{basedir}/{library_id}_{filtering_method}_processed_sce.rds"
+        processed_sce =  os.path.join(config['results_dir'], "{sample_id}/{library_id}_{filtering_method}_processed_sce.rds")
     output:
-        "{basedir}/{library_id}_{filtering_method}_core_analysis_report.html"
+        os.path.join(config['results_dir'], "{sample_id}/{library_id}_{filtering_method}_core_analysis_report.html")
```
jashapiro, line 127: I think you will want to leave these with `{basedir}`, but you will then update the `output_dir` below to `output_dir = '{wildcards.basedir}/{wildcards.sample_id}'

cluster.snakefile
```diff
@@ -0,0 +1,65 @@
...
+
+
+rule calculate_clustering:
+    input:
+        get_input_rds_files
+    output:
+        "{base_dir}/{sample_id}/{library_id}_processed_sce_clustering.rds",
+        "{base_dir}/{sample_id}/clustering_stats/{library_id}_clustering_all_validity_stats.tsv",
+        "{base_dir}/{sample_id}/clustering_stats/{library_id}_clustering_summary_validity_stats.tsv",
+        "{base_dir}/{sample_id}/clustering_stats/{library_id}_clustering_summary_stability_stats.tsv"
```
jashapiro, line 51: for consistency, `base_dir` -> `basedir` throughout.

cluster.snakefile
```diff
@@ -0,0 +1,65 @@
...
+               library = LIBRARY_ID),
+        expand(os.path.join(config["results_dir"], "{sample}/clustering_stats/{library}_clustering_summary_stability_stats.tsv"),
+               zip,
+               sample = SAMPLES,
+               library = LIBRARY_ID)
+
+
+def get_input_rds_files(wildcards):
+    lib_info = samples_information.set_index('library_id')
+    return lib_info.loc[wildcards.library_id]['filepath']
```
jashapiro, line 41: If I am understanding correctly, I think you will want this to be something like: 

```suggestion
def get_input_rds_files(wildcards):
    lib_info = samples_information.set_index('library_id')
    filter_method = lib_info.loc[wildcards.library_id]['filtering_method']
    path = f"{wildcards.base_dir}/{wildcards.sample_id}/{wildcards.library_id}_{filter_method}_processed_sce.rds"
```

So we are using the info file to figure out what the filtering method should be and then filling that in for the file input

Note that this means the filtering method will be lost from the output. If we want to keep it, we can incorporate the filter method into the `target` rule creation, then we might be able use just:
```
input:
    "{base_dir}/{sample_id}/{library_id}_{filter_method}_processed_sce.rds"
output:
    "{base_dir}/{sample_id}/{library_id}_{filter_method}_clustered_sce.rds"
```

(Note I shortened to `_clustered_sce.rds` rather than just adding an appendix)

allyhawkins (commented):
I just want to note that I agree with Josh's comments about changing the output to the rule to be just the clustering stats directory and then keeping the filtering method in the name as mentioned in https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/214#discussion_r970014384. If you do that you should be able to completely remove the `def get_input_rds_files` function. Once those changes are made this should be good to go.

jashapiro:
> I just want to note that I agree with Josh's comments about changing the output to the rule to be just the clustering stats directory and then keeping the filtering method in the name as mentioned in [#214 (comment)](https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/214#discussion_r970014384). If you do that you should be able to completely remove the `def get_input_rds_files` function. Once those changes are made this should be good to go.

In that case, we will need to make `clustering_stats/{library_id}` subdirectories.

allyhawkins:
> In that case, we will need to make clustering_stats/{library_id} subdirectories.

Why is that? Couldn't you do the following? 

```
input:
    "{base_dir}/{sample_id}/{library_id}_{filter_method}_processed_sce.rds"
output:
    "{base_dir}/{sample_id}/{library_id}_{filter_method}_clustered_sce.rds",
    "{base_dir}/{sample_id}/clustering_stats"
```

cbethell:
> > In that case, we will need to make clustering_stats/{library_id} subdirectories.
> 
> Why is that? Couldn't you do the following?
> 
> ```
> input:
>     "{base_dir}/{sample_id}/{library_id}_{filter_method}_processed_sce.rds"
> output:
>     "{base_dir}/{sample_id}/{library_id}_{filter_method}_clustered_sce.rds",
>     "{base_dir}/{sample_id}/clustering_stats"
> ```

The above would result in the error I mentioned as part of number 1 in https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/214#issuecomment-1245841206, snakemake is looking for `{library_id}` in each of the specified output paths.

allyhawkins:
Ahhh okay... What about doing the following to minimize the number of nested folders with only one folder? Especially because many samples only have one library: 

```
input:
    "{base_dir}/{sample_id}/{library_id}_{filter_method}_processed_sce.rds"
output:
    "{base_dir}/{sample_id}/{library_id}_{filter_method}_clustered_sce.rds",
    "{base_dir}/{sample_id}/{library_id}_{filter_method}_clustering_stats"
```

cbethell:
> Ahhh okay... What about doing the following to minimize the number of nested folders with only one folder? Especially because many samples only have one library:
> 
> ```
> input:
>     "{base_dir}/{sample_id}/{library_id}_{filter_method}_processed_sce.rds"
> output:
>     "{base_dir}/{sample_id}/{library_id}_{filter_method}_clustered_sce.rds",
>     "{base_dir}/{sample_id}/{library_id}_{filter_method}_clustering_stats"
> ```

This would mean removing this chunk from the calculations Rscript (which is fine with me), just want to make sure this is fine with other folks:

```
stats_output_dir <- file.path(opt$output_directory, "clustering_stats")
if(!dir.exists(stats_output_dir)){
  dir.create(stats_output_dir, recursive = TRUE)
}
```

jashapiro:
That should work, with a small update:
```
input:
    "{base_dir}/{sample_id}/{library_id}_{filter_method}_processed_sce.rds"
output:
    "{base_dir}/{sample_id}/{library_id}_{filter_method}_clustered_sce.rds",
    directory("{base_dir}/{sample_id}/{library_id}_{filter_method}_clustering_stats")
```

If you aren't using the `sample_id` this could potentially be simplified further with fewer wildcards, but I don't think they hurt.

allyhawkins:
> Would also mean that the RDS file will be saved within this clustering stats directory

I don't believe this is the case. The RDS file should still be saved just within `{basedir}/{sample_id}` and then there will be a separate stats directory for each library within that directory.

cbethell:
> > Would also mean that the RDS file will be saved within this clustering stats directory
> 
> I don't believe this is the case. The RDS file should still be saved just within `{basedir}/{sample_id}` and then there will be a separate stats directory for each library within that directory.

This is the output file structure I get without making the change to the calculations Rscript:

Screen Shot 2022-09-14 at 2 33 35 PM

I believe since we manually define "clustering_stats" as a directory in the script, we are still getting that subdirectory, but on the same level as the RDS file because we are using the `out_dir` in the snakefile to define the output directory for the RDS file (which was fine when we manually defined the clustering_stats subdirectory in the R script).

allyhawkins:
> I believe since we manually define "clustering_stats" as a directory in the script, we are still getting that subdirectory, but on the same level as the RDS file because we are using the out_dir in the snakefile to define the output directory for the RDS file (which was fine when we manually defined the clustering_stats subdirectory in the R script, if that makes sense).

Okay I see what you are saying. I thought you meant that the snakefile was putting them in the same directory which should not be the case. Definitely on board with your changes to remove the definition of the `clustering_stats` directory within the R script and explicitly provide the output path as an argument.

config.yaml
```diff
@@ -32,3 +32,7 @@ n_genes_pca: 2000 # The `n` number of highly variable genes to subset for PCA
 # clustering parameters
 cluster_type: "louvain"
 nearest_neighbors: 10
+nearest_neighbors_min: 5 # The minimum value to use for a range of neareast neighbors values for exploration
+nearest_neighbors_max: 25 # The maximum value to use for a range of neareast neighbors values for exploration
+nearest_neighbors_increment: 5 # The increment to use when implementing the range number of nearest neighbors for cluster stats
+overwrite_results: TRUE # A binary value indicating whether or not to overwrite any existing clustering results
```
allyhawkins, line 38:
```suggestion
overwrite_results: TRUE # A binary value indicating whether or not to overwrite any existing clustering results

```

allyhawkins (approved):
LGTM 🚀

jashapiro (approved):
With the caveat that I have not tried to run this myself, this looks good to me. 

So if it runs as expected for you, I expect it to work!

PR #215, opened 2022-09-09
Optional clustering Snakefile - rendering template report rule

### LINKED ISSUE #209: Initiate Snakefile for clustering module - rendering template notebook rule

Once the optional clustering template notebook (https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/207) and the first rule of the Snakefile (#180) are completed, we will want to add a second rule to the Snakefile in this optional module to render the report template notebook.

Once there is a Snakefile in the optional-clustering-analysis module, the second rule to be added will be a `render-report` rule that renders the clustering report template notebook to generate a html file, using the output of the first rule (the clustering calculations rule).

We will likely want to do this similarly to the core report template rendering rule (found in this repo's main Snakefile), where we take the output of the first rule and allow users to define where to save the rendered html file.

cbethell:
**Issue Addressed**
Closes #209

This PR adds the second rule of the `cluster.snakefile` necessary to complete the optional clustering analysis.

This PR adds a rule named `generate_cluster_report` to run the `clustering-report-template.Rmd` to generate a clustering report associated with each `_processed_sce_clustering.rds` file (the output of the first calculations rule of the snakefile.

Again, as noted in https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/214#issuecomment-1242117804 on the PR that this is stacked on, the template notebook is giving an error when I attempt to run `snakemake --snakefile cluster.snakefile -c2 --use-conda` so I left this in the root directory _for now_. See the error for reference:
Screen Shot 2022-09-09 at 9 41 59 AM

- Does anyone have thoughts on a solution to the error above?
- I also played around with the input/output paths in `cluster.snakefile` to get things to run successfully, so I am also open to refactoring suggestions there!

**Checklist**


cluster.snakefile
```diff
@@ -63,3 +67,26 @@ rule calculate_clustering:
...
+    output:
+        os.path.join(config['results_dir'], "{sample_id}/{library_id}_optional_clustering_report.html")
+    conda: "envs/scpca-renv.yaml"
+    shell:
+        """
+        R_PROFILE_USER='.Rprofile' \
+        Rscript -e \
+        "rmarkdown::render('{workflow.basedir}/clustering-report-template.Rmd', \
+                           clean = TRUE, \
+                           output_file = '{output}', \
```
allyhawkins, line 83:
```suggestion
        "rmarkdown::render('{workflow.basedir}/optional-clustering-analysis/clustering-report-template.Rmd', \
                           clean = TRUE, \
                           output_file = '{output}', \
                           output_dir = dirname('{output}') \
```
I believe this should help solve the issue by directing the output. Also `rmarkdown::render()` has this weird thing I discovered recently where it doesn't like to make the directory of the file it's writing out so it will throw the error that it doesn't exist unless you specify the `output_dir`. If you specify the directory then it will make the directory. Either way we want to make sure we are telling the notebook to write the output to the full results path, not the path relative to where the notebook is running.

jashapiro (commented):
My suggestions here mostly follow from the suggestions in #214. Let snakemake handle most of the path finding, and take advantage of using a directory for the output of the clustering stats to make that a second input to the report step. This makes the dependencies a bit more explicit while saving places that might need changes downstream.

cbethell:
> My suggestions here mostly follow from the suggestions in #214. Let snakemake handle most of the path finding, and take advantage of using a directory for the output of the clustering stats to make that a second input to the report step. This makes the dependencies a bit more explicit while saving places that might need changes downstream.

Took your suggestion here @jashapiro and also propagated another change over from #214 👍

allyhawkins (approved):
These changes look good to me and everything ran as expected on my computer 👍

jashapiro (approved):
I realized that I had already made some comments, but never submitted them as a review. Sorry about that!

This looks good to me, with my only major comment being that we can get rid of the "optional" part throughout. I expect the clustering to merge into the main workflow before too long, so to save editing later, we could remove that now. 

I'm approving, but if you do modify file and directory names, make sure you test that everything still works.

cbethell:
> This looks good to me, with my only major comment being that we can get rid of the "optional" part throughout. I expect the clustering to merge into the main workflow before too long, so to save editing later, we could remove that now.
> 
> I'm approving, but if you do modify file and directory names, make sure you test that everything still works.

Thanks for the feedback @jashapiro! I wanted to make a note here that I implemented all of your suggestions with exception of modifying the directory name because I am not yet sure if we _are_ going to merge the clustering stuff into the main workflow (with the initial idea being that this module/directory would be one of numerous _optional_ modules that folks can implement after running the main workflow).

PR #216, opened 2022-09-09
Fix restore argument to use input lockfile

jashapiro:
In https://github.com/AlexsLemonade/sc-data-integration/pull/136#issuecomment-1242485543, I noticed that the lockfile was mis-specified in the `build_renv` rule, which seems to make rebuilding when the `renv.lock` file changes fail with the statement that the lock file is up to date. This fixes the arguments, and should hopefully make updating work the way it was expected to.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -44,7 +44,7 @@ rule build_renv:
     shell:
       """
       R_PROFILE_USER='{workflow.basedir}/.Rprofile' \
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
Thanks for catching this and for making this change! Looks good

jashapiro:
> Thanks for catching this and for making this change! Looks good

Well, you caught it! I just figured out the why.

PR #219, opened 2022-09-13
Pin pandoc version

jashapiro:
This is a small PR to avoid a potential problem where a version of pandoc (2.19.1) is broken resulting in bad rendering of Rmd files.

Note that this will require rebuilding the renv environment, which can now be done with `bash setup_envs.sh` (or `./setup_envs.sh` if you want to save keystrokes).


sjspielman (approved):
👍

PR #222, opened 2022-09-23
Add logs to snakemake rules

jashapiro:
**Issue Addressed**
N/A

This PR adds `log:` directives to the Snakefiles in this repository. I originally was prompted to do this because I rand `snakemake --lint` as an thought for testing, and saw that this was recommended, and it comes with a few advantages: it should make debugging rules a bit easier, as the logs are now written to files rather than just the screen, and this also means that the screen output is _only_ snakemake rules, which saves a bit of random text & prevents multiple jobs from writing to the screen at once.

The primary change is adding the `log:` directive to all rules, which will store files in the `logs/` directory, arranged similarly to the input data. The log file path has to include all wildcards, hence the nested directory structure. I could alternatively have made very long names for the log files, but I thought nesting was just as easy. Happy to reconsider this though.

Each rule also adds a line for outputting stdout and stderr to the log file, usually in the form `&> {log}`.

I also did some cleanup removing no longer needed references to `{workflow.basedir}` and `.Rprofile` as those should now be covered by always running the workflows from the main project directory. 

Not really: just whether to change the output paths, but doing it another way would mean breaking up some of the `{basedir}` wildcards, which seemed like trouble.

`snakemake --lint` also complains about using the config settings directly in shell commands. It would prefer that they be defined as `params` and used that way. I decided not to do that here, as it adds a lot of boilerplate, and doesn't give as obvious a benefit for debugging/cleaner screens.

Info about logs should probably be added to the docs, but I don't think great detail is needed.

**Checklist**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
 # ignore the `data`, `results`, and `example-results` directories and the files within
-/data/
-/results/
-/example-results/
+data/
+results/
+example-results/
+logs/
 
 # ignore hidden `DS_Store`
 .DS_Store
--- a/Snakefile
+++ b/Snakefile
@@ -40,11 +40,12 @@ def get_input_rds_files(wildcards):
 rule build_renv:
     input: workflow.source_path("renv.lock")
     output: "renv/.snakemake_timestamp"
+    log: "logs/build_renv.log"
     conda: "envs/scpca-renv.yaml"
     shell:
       """
       R_PROFILE_USER='{workflow.basedir}/.Rprofile' \
-      Rscript -e "renv::restore(lockfile = '{input}')"
+      Rscript -e "renv::restore(lockfile = '{input}')" &> {log}
       date -u -Iseconds  > {output}
       """
 
@@ -53,7 +54,8 @@ rule filter_data:
     input:
         get_input_rds_files
     output:
-        temp(os.path.join(config['results_dir'], "{sample_id}/{library_id}_{filtering_method}_filtered.rds"))
+        temp("{basedir}/{library_id}_{filtering_method}_filtered.rds")
+    log: "logs/{basedir}/{library_id}_{filtering_method}/filter_data.log"
     conda: "envs/scpca-renv.yaml"
     shell:
         "R_PROFILE_USER='{workflow.basedir}/.Rprofile'"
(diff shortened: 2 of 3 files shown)


cbethell (commented):
Great idea to include the logs here @jashapiro, thanks!

I now encounter a pandoc error when I attempt to run the snakefile with the changes in this PR however. Do you think this is possibly due to some pandoc update?

Screen Shot 2022-09-23 at 9 58 52 AM

jashapiro:
That is a conda error, which should not be related to any changes that I made in this PR. We pinned the pandoc version a few revisions ago because of a bug in one of the later version of pandoc, but I am not sure why it thinks it can't download it here.

jashapiro:
Jus to check I just rebuilt my environments in the `scpca-downstream-analysis` project with `./setup_envs.sh`, and everything seemed to install correctly. I'm not sure why your conda might not have been able to find pandoc 2.18.

jashapiro:
Ah, looking at https://anaconda.org/conda-forge/pandoc/files, it seems like for some reason the `osx-arm64` builds of pandoc don't exist, except for version 2.19. Which makes me think that snakemake is trying to rebuild the environment because something may have changed since the last time your ran the workflow, but it is failing because it is not using the Intel build environment. Rerunning `setup_envs.sh` _should_ solve the problem.

cbethell:
> 

Unfortunately I am still getting the pandoc error linked in https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/222#pullrequestreview-1118636195, but this now convinces me that it is indeed some `osx-arm64` error.

jashapiro:
> > Ah, looking at https://anaconda.org/conda-forge/pandoc/files, it seems like for some reason the osx-arm64 builds of pandoc don't exist, except for version 2.19. Which makes me think that snakemake is trying to rebuild the environment because something may have changed since the last time your ran the workflow, but it is failing because it is not using the Intel build environment. Rerunning setup_envs.sh should solve the problem.
> 
> Unfortunately I am still getting the pandoc error linked in [#222 (review)](https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/222#pullrequestreview-1118636195), but this now convinces me that it is indeed some `osx-arm64` error.

Can you try deleting the `.snakemake/conda` directory and rerunning `setup_envs.sh`? If it is working correctly, you should not be creating any `osx-arm64` environments. If it doesn't work, ping me on Slack so we can do some debugging. We'd really like to keep pandoc pinned if we can.

cbethell (approved):
> Can you try deleting the .snakemake/conda directory and rerunning setup_envs.sh? If it is working correctly, you should not be creating any osx-arm64 environments. If it doesn't work, ping me on Slack so we can do some debugging. We'd really like to keep pandoc pinned if we can.

As we discovered this morning, the issue is solved by running things on my local Mac terminal rather than RStudio terminal (as you noted @jashapiro, RStudio needs to complete the Electron transition before the `setup_envs.sh` script can work as expected there).

That said I am going to approve this PR now that I can confirm things work here ✅

Issue #218, opened 2022-09-13
Add detailed docs on the clustering workflow options/decisions


cbethell:
**Please provide some background on the proposed additions or changes.**

Once we have accounted for the possible use of more than one `cluster_type` when running the optional clustering workflow, we will want to update the optional clustering module's documentation.

**What are the changes that you are proposing?**

Updates to the documentation should include options that folks can implement when running the workflow (e.g. you can specify `walktrap` and `louvain` to test a range of parameters for both types of clustering). We should also be clear about any decisions made here (e.g. any values added to the config file, etc.).

This is also likely when we would want to include or expand on expected inputs and outputs sections.

cbethell:
This has been addressed by merged PR #228

PR #224, opened 2022-09-28
Account for additional `cluster_type` in clustering calculations script

### LINKED ISSUE #212: Account for additional `cluster_type` in optional clustering - calculations script

As discussed in a check in meeting, we may want to think about accounting for an additional `cluster_type` in the optional clustering module once we have completed the initial workflow using just one cluster type with multiple nearest neighbors values.
In other words, we will want to allow for `louvain` and `walktrap` clustering with the specified nearest neighbors range of values.

This change will require us to add a feature to the `clustering_calculations.R` script, as well as the template notebook, which will allow the user to specify two `cluster_type`(s), rather than one. 
**Edit:** However this issue should track just updates to the `clustering_calculations.R` script (per the comment left below).

Initial thoughts here, we will want to be able to perform two sets of clustering and associated cluster validity/stability calculations on each of the specified cluster types, but still also allow for only one type of clustering if chosen by the user. This will likely mean an if statement of some sort.

Not specifically but an if statement handling two versus one clustering type can get a bit tricky.

cbethell:
**Issue Addressed**
Closes #212 

This PR adds an additional `cluster_type` parameter to the `optional-clustering-analysis/clustering-calcutions.R` script, as to allow for users to compare more than one `cluster_type` across the same range of parameters.

More specifically, this PR:

1. Adds an optparse param to the `clustering-calculations.R` script named `second_cluster_type` 
2. Checks if there is a second cluster type provided and performs clustering for that type if it provided
3. Creates wrapper functions to check the cluster validity and stability stats, as to have all of the cluster stats (for all specified cluster types) in each relevant output file

This PR also removes the 2.18 specification associated with the pandoc required dependency, due to the following error and also related to https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/222#pullrequestreview-1118636195:

Screen Shot 2022-09-23 at 9 58 52 AM

Removing the "2.18" allowed things to run successfully.

**Edit:** mostly fixed the above error so reverted the pandoc change here

- Does the code added in this PR seem like the most suitable way to tackle #212? Do you have suggestions for refactoring?
- 
**Checklist**

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -31,7 +31,7 @@ n_genes_pca: 2000 # The `n` number of highly variable genes to subset for PCA
 
 # clustering parameters
 cluster_type: "louvain"
-cluster_types: "louvain,walktrap"
+optional_cluster_types: "louvain,walktrap"
 nearest_neighbors: 10
 nearest_neighbors_min: 5 # The minimum value to use for a range of neareast neighbors values for exploration
 nearest_neighbors_max: 25 # The maximum value to use for a range of neareast neighbors values for exploration
--- a/config.yaml
+++ b/config.yaml
@@ -31,6 +31,7 @@ n_genes_pca: 2000 # The `n` number of highly variable genes to subset for PCA
 
 # clustering parameters
 cluster_type: "louvain"
+cluster_types: "louvain,walktrap"
 nearest_neighbors: 10
 nearest_neighbors_min: 5 # The minimum value to use for a range of neareast neighbors values for exploration
 nearest_neighbors_max: 25 # The maximum value to use for a range of neareast neighbors values for exploration
--- a/optional-clustering-analysis/clustering-calculations.R
+++ b/optional-clustering-analysis/clustering-calculations.R
@@ -11,8 +11,7 @@
 #   --sce "example-results/sample01/library01_miQC_processed_sce.rds" \
 #   --library_id "library01" \
 #   --seed 2021 \
-#   --cluster_type "louvain" \
-#   --second_cluster_type "walktrap" \
+#   --cluster_types "louvain,walktrap" \
 #   --nearest_neighbors_min 5
 #   --nearest_neighbors_max 25 \
 #   --nearest_neighbors_increment 5 \
@@ -49,16 +48,10 @@ option_list <- list(
     metavar = "integer"
   ),
   optparse::make_option(
-    c("-c", "--cluster_type"),
+    c("-c", "--cluster_types"),
     type = "character",
-    default = "louvain",
+    default = "louvain,walktrap",
     help = "Method used for clustering. Can be either louvain or walktrap.",
-  ),
-  optparse::make_option(
-    c("--second_cluster_type"),
-    type = "character",
-    default = NULL,
-    help = "Additional method used for clustering. Can be either louvain or walktrap.",
   ),
   optparse::make_option(
     c("-n", "--nearest_neighbors_min"),
@@ -145,9 +138,14 @@ if (!file.exists(opt$sce)){
   stop(paste(opt$sce, "does not exist."))
 }
 
-# Check that clustering type is valid 
-if (!opt$cluster_type %in% c("louvain", "walktrap")) {
-  stop("--cluster_type (-c) must be either louvain or walktrap.")
+# Split up string of cluster types
+cluster_types <- unlist(stringr::str_split(opt$cluster_types, ","))
+
+# Check that clustering type is valid
+for (cluster_type in cluster_types) {
+  if (!cluster_type %in% c("louvain", "walktrap")) {
+    stop("--cluster_types (-c) must be either louvain or walktrap.")
+  }
 }
 
 # Make sure the output directory exists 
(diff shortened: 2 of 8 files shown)


optional-clustering-analysis/clustering-calculations.R
```diff
@@ -169,47 +176,128 @@ nn_range <- define_nn_range(opt$nearest_neighbors_min,
...
+    second_summary_stability_stats_df <- 
+      get_cluster_stability_summary(sce,
+                                    nn_range,
+                                    second_cluster_type) %>%
+      dplyr::mutate(param_value = as.numeric(param_value))
+    
+    summary_stability_stats_df <- summary_stability_stats_df %>%
+      dplyr::bind_rows(second_summary_stability_stats_df)
+  }
+}
```
allyhawkins, line 295:
```suggestion
  # Check the cluster stability stats and return the summary ari in a data frame
  summary_stability_stats_df_list <- purrr::map(cluster_types, 
    ~ get_cluster_stability_summary(sce,
                                  nn_range,
                                  .x )
 summary_stability_stats_df <- dplyr::bind_rows(summary_stability_stats_df) &>&
    dplyr::mutate(param_value = as.numeric(param_value))
 
```
Rather than have a whole other function you can condense this and use `purrr` here to make a list of dataframes, one for each cluster type and then combine together into one dataframe.

optional-clustering-analysis/clustering-calculations.R
```diff
@@ -169,47 +176,128 @@ nn_range <- define_nn_range(opt$nearest_neighbors_min,
...
+
+# Run the cluster validity stats wrapper function
+validity_stats_df <- cluster_validity_stats_wrapper_function(
+  sce, nn_range, opt$cluster_type, 
+  second_cluster_type = opt$second_cluster_type)
```
allyhawkins, line 300:
```suggestion
```

allyhawkins (commented):
Thanks for getting this started. My overall comment is to actually use a single argument instead of two and read in a string of cluster options and separate that into a vector. You can then loop through that vector of cluster options and apply the functions that you have written. This should hopefully simplify things and also make things more robust to any future changes like additional clustering we would want to do.

cbethell:
Thanks for the helpful feedback @allyhawkins! I was able to revert back to one `cluster_type` param option, supply the vector of multiple cluster types there, and use `str_split()` to handle those cluster types individually. I also added the line to the Snakefile in the last commit https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/224/commits/c6b2f0ae7f71f5e121ee8c6246e82f9705e52ae6 that gets the modifications in this PR working successfully. Noting that I added a `cluster_types` parameter to the config.yaml file since we use `cluster_type` in the core workflow and only expected one cluster type there -- let me know what you think here and overall!

allyhawkins (approved):
LGTM!

config.yaml
```diff
@@ -31,6 +31,7 @@ n_genes_pca: 2000 # The `n` number of highly variable genes to subset for PCA
 
 # clustering parameters
 cluster_type: "louvain"
```
sjspielman, line 33: This line should be removed now, right?

optional-clustering-analysis/clustering-calculations.R
```diff
@@ -138,9 +138,14 @@ if (!file.exists(opt$sce)){
   stop(paste(opt$sce, "does not exist."))
 }
 
-# Check that clustering type is valid 
-if (!opt$cluster_type %in% c("louvain", "walktrap")) {
-  stop("--cluster_type (-c) must be either louvain or walktrap.")
+# Split up string of cluster types
+cluster_types <- unlist(stringr::str_split(opt$cluster_types, ","))
```
sjspielman, line 142: If you want to protect against someone adding a space, e.g.`louvain, walktrap`, we can do:


```suggestion
cluster_types <-  stringr::str_split(opt$cluster_types, ",") %>% unlist() %>% stringr::str_trim()
```

optional-clustering-analysis/clustering-calculations.R
```diff
@@ -138,9 +138,14 @@ if (!file.exists(opt$sce)){
...
+
+# Check that clustering type is valid
+for (cluster_type in cluster_types) {
+  if (!cluster_type %in% c("louvain", "walktrap")) {
+    stop("--cluster_types (-c) must be either louvain or walktrap.")
```
sjspielman, line 147: Update this message - it can be either or BOTH, comma separated.

optional-clustering-analysis/clustering-calculations.R
```diff
@@ -169,47 +174,64 @@ nn_range <- define_nn_range(opt$nearest_neighbors_min,
                             opt$nearest_neighbors_max,
                             opt$nearest_neighbors_increment)
 
-# Check for existing clustering results
-cluster_column_names <- paste(opt$cluster_type, nn_range, sep = "_")
-existing_columns <- intersect(cluster_column_names, colnames(colData(sce)))
-
-if(length(existing_columns) != 0){
+cluster_wrapper_function <- function(sce, nn_range, cluster_type, ...) {
```
sjspielman, line 177: I generally like for function names to be more action-oriented, which is a preference I get from the style guide: https://style.tidyverse.org/functions.html#naming 

In this case, mostly I don't like that `function` is in the name of the function. We know it's a function, but what does it do? It performs the clustering, and incidentally does so by wrapping other functions. 
How do you feel about naming it (and updating its use below) `perform_clustering()`?

config.yaml
sjspielman: Ah, I see. The names are a bit to close for comfort. Seems really easy to use the wrong one. How about `cluster_types` --> `opt_cluster_types` (like "optional")? I'm not sure about that either though. Tagging @allyhawkins back in for ideas!

allyhawkins: Hmm looking at this I have a few ideas. Both of which would include changing the name. I would say maybe we need to change both argument names so that we have `core_cluster_type` and `optional_cluster_types` or something like that? So that we are more explicit? 

Additionally I think we can do some reorganization but that could go one of two ways.  My first ideas is to actually split out any of the clustering specific parameters into their own config file that are parameters that just get accessed in this clustering exploration module. Then you can call both config files at the beginning of the snakefile with something like this (assuming that all the config files also get moved to a new config directory which I would also suggest).
```
configfile = "config/config.yaml"
configfile = "config/cluster_config.yaml"
```

The other option is to just move all the optional clustering parameters to another section in the larger main configfile with a comment above explaining that they are for that module. We want to make sure that users understand where each parameter goes. 

I think I favor the first option because someone using just the core workflow might be confused as to why there are these extra parameters that are irrelevant to what they are using. Additionally we are going to have more modules in the future so we will probably want to have config files that have module specific parameters moving forward.

sjspielman: I like `core_cluster_type` and `optional_cluster_types` as explicit parameter names.

Just gonna also note that I find that `Snakemake` behavior WEIRD. I would expect writing `configfile` twice to overwrite it, but apparently not ¯\_(ツ)_/¯

allyhawkins: >Just gonna also note that I find that Snakemake behavior WEIRD. I would expect writing configfile twice to overwrite it, but apparently not 🤷‍♀️

It would overwrite it if you had conflicting parameter names I think, but if they are unique parameters then it should use all of them as far as I can tell in the docs.

sjspielman: that makes some sense I suppose..

cbethell:
Thank you both for the feedback @sjspielman and @allyhawkins! I have created separated out the parameters to have two config files as Ally suggested, renamed the parameters to be more explicit, and successfully tested out the changes! Let me know what you think now!

config.yaml
jashapiro: I'll weigh in with a slight difference: I would leave `config.yaml` in root directory, and only move `cluster_config.yaml` to the subdirectory, which I might name `opt_configs` or something like that.

And a note on syntax, which might make @sjspielman a bit happier:
```
configfile: "config.yaml"
configfile: "opt_configs/cluster_config.yaml"
```

(Note the colon instead of `=`)

cbethell: As Josh and I discussed in science pair it, I have made personal note of his comment above in case it comes up in discussion later, but will leave things as is for now!

cbethell:
> Changes look good to me! The only other suggestion I have is to update the README with this new parameter setup. There are several places where this should be changed throughout.

Sweet, thank you! I am planning to put the README updates in a subsequent PR -- likely the same PR that will be tackling https://github.com/AlexsLemonade/scpca-downstream-analyses/issues/218!

sjspielman (approved):
👍

PR #226, opened 2022-10-07
Account for additional `cluster_type` in clustering report

### LINKED ISSUE #217: Account for additional cluster_type in optional clustering - template notebook

As noted on issue #212, we want to account for more than one graph-based `cluster_type` for users to test in the optional clustering module.
In other words, we will want to allow for `louvain` and `walktrap` clustering with the specified nearest neighbors range of values.

This issue should track the changes to be made to the optional clustering report template Rmd file. This would be allowing for two sets of plots only if a second `cluster_type` is provided.

Perhaps we can include a check in the notebook that would determine if there is a second `cluster_type` provided, and plot a second set of the existing plots for that `cluster_type` (perhaps may mean implementing `purrr::map()` or a for loop)

None that I can think of at this time.

cbethell:
**Issue Addressed**
Closes #217

Now that the modification that allows for more than one type of clustering has been made to the clustering calculations script in PR #224, we want to account for the additional cluster type in the clustering report template notebook as well.

This PR does the following:
- implements `sprintf()` where we create the cluster column names to fix the order of plots when displayed
- adds a for loop to the `clustering-report-template.Rmd` file to loop through a possible comma-separated list of more than one cluster type
- modifies the underlying plotting functions in `clustering-functions.R` to allow for plotting of more than one cluster type

- Does this PR seem to satisfy #217?
- Any suggestions on how the plots, comparing the parameters for each cluster type, should be displayed?

See an example html report with the changes made in this PR for reference:
[library01_miQC_clustering_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/9748258/library01_miQC_clustering_report.html.zip)

**Checklist**

### DIFF
--- a/optional-clustering-analysis/clustering-report-template.Rmd
+++ b/optional-clustering-analysis/clustering-report-template.Rmd
@@ -165,17 +165,25 @@ knitr::opts_chunk$set(
 
 ```{r}
 # Plot UMAP coordinates
-plot_list <- existing_columns %>%
-  purrr::map(
-    ~ plotReducedDim(processed_sce, dimred = "UMAP",colour_by = .x,
-                     point_size = 0.4, point_alpha = 0.5) +
-      theme_bw() +
-      theme(text = element_text(size = 14),
-            legend.text = element_text(size = 14)) +
-      guides(color = guide_legend(override.aes = list(size = 1)))
-  )
-
-cowplot::plot_grid(plotlist = plot_list, ncol = num_col)
+for (cluster_type in cluster_types) {
+  existing_columns_filtered <- existing_columns[stringr::str_detect(existing_columns, cluster_type)]
+  plot_list <- existing_columns_filtered %>%
+    purrr::map(
+      ~ plotReducedDim(
+        processed_sce,
+        dimred = "UMAP",
+        colour_by = .x,
+        point_size = 0.4,
+        point_alpha = 0.5
+      ) +
+        theme_bw() +
+        theme(text = element_text(size = 14),
+              legend.text = element_text(size = 14)) +
+        guides(color = guide_legend(override.aes = list(size = 1)))
+    )
+  
+  cowplot::plot_grid(plotlist = plot_list, ncol = num_col)
+}
 ```
 
 ## Evaluating clustering
--- a/optional-clustering-analysis/clustering-report-template.Rmd
+++ b/optional-clustering-analysis/clustering-report-template.Rmd
@@ -64,10 +64,13 @@ library(ggpubr)
 library(scater)
 library(dplyr)
 
+# Split up string of cluster types
+cluster_types <- unlist(stringr::str_split(params$cluster_type, ","))
+
 # Set theme for plotting
 library(ggplot2)
-theme_set(theme_bw() + theme(text = element_text(size = 17),
-                             legend.text = element_text(size = 16)))
+theme_set(theme_bw() + theme(text = element_text(size = 17*length(cluster_types)),
+                             legend.text = element_text(size = 16*length(cluster_types))))
 point_size <- params$point_size
 ```
 
@@ -83,9 +86,6 @@ nn_range <- define_nn_range(params$nearest_neighbors_min,
                             params$nearest_neighbors_increment)
 
 cluster_column_names <- c()
-
-# Split up string of cluster types
-cluster_types <- unlist(stringr::str_split(params$cluster_type, ","))
 
 # Grab column name with clustering results
 for (cluster_type in cluster_types) {
@@ -154,8 +154,13 @@ Here clustering was performed using `r params$cluster_type` clustering.
 ```{r}
 # Set number of columns for plotting and use this to define figure heights 
 # and widths
-num_col <- 2
-figure_dim <- max(ceiling(length(existing_columns)/num_col) * 5, 4)
+if (length(cluster_types) == 1) {
+  num_col <- 2
+} else {
+  num_col <- length(cluster_types)
+}
+
+figure_dim <- max(ceiling(length(existing_columns)/num_col) * length(existing_columns)/num_col + 2, (length(existing_columns)/num_col) + 1)
 
 knitr::opts_chunk$set(
   fig.height = figure_dim,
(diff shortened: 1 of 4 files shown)


utils/clustering-functions.R
```diff
@@ -392,9 +391,8 @@ plot_cluster_silhouette_width <- function(cluster_validity_df, num_col, point_si
     ggplot(metadata, aes(x = cluster, y = width)) +
     ggbeeswarm::geom_quasirandom(method = "pseudorandom", size = point_size) +
     geom_hline(yintercept = 0, linetype = 'dotted') +
-    labs(title = unique(metadata$cluster_type),
-         x = "Cluster Assignment") +
-    facet_wrap( ~ param_value, scale="free_x", ncol = num_col) +
+    labs(x = "Cluster Assignment") +
+    facet_wrap( ~ cluster_names_column, scale="free_x", ncol = num_col) +
```
allyhawkins, line 395:
```suggestion
    facet_wrap( ~ cluster_names_column, scale="free_x", ncol = num_col, dir = "v" ) +
```
I think we want to add this here to so that we can have a column for each cluster type and then a row for each `nn_param` tested.

allyhawkins (commented):
A few general comments: 

- I think we want to somehow group the UMAPs. Either we want to plot them like the other UMAPs so that we can directly compare louvain nn=5 to walktrap nn=5 where each of them is in a column or we want to make UMAPs for each clustering type and plot them separately. I do like being able to directly compare them. 
- All of the plots where you are faceting by cluster/nn_param combo should be the same format as the purity plots where you have `louvain_10` and `walktrap_10` in the same row. In doing that you might want to alter the number of columns used for plotting based on the `length(cluster_types)` used. I would also check that if the length is only 1, you still use 2 as your default.
- Across the board the size of fonts still look kind of small, can we increase those?

sjspielman:
> I think we want to somehow group the UMAPs. Either we want to plot them like the other UMAPs so that we can directly compare louvain nn=5 to walktrap nn=5 where each of them is in a column or we want to make UMAPs for each clustering type and plot them separately. I do like being able to directly compare them.

+1 ^^. The current layout is a bit tricky to see/read, and font size is also part of that.

cbethell:
Thank you both for the feedback! Things are now being compared based on the param range, and the font sizes have been adjusted to be less-hardcoded and more dynamic based on the number of cluster types! The UMAP font sizes still seem a bit small to me but if I make them any bigger some of the legend is cut-off -- let me know if you have any suggestions there and any further feedback!

allyhawkins (commented):
Can you update the legends for the purity and the summary plots to have larger points and lines? They are hard to see. 

The UMAPs look better to me now that they are paired but I wonder if we could label each column so that they have the type of clustering listed as the title for that column. I think that to do that you would have to break up how you are creating the plots and `purrr::map` through the `cluster_types` and then combine the two columns which might get complicated. The font size also is still a little small, but I know you need to fit them all in so I don't know what else we can do about that.

cbethell:
> Can you update the legends for the purity and the summary plots to have larger points and lines? They are hard to see.
> 
> The UMAPs look better to me now that they are paired but I wonder if we could label each column so that they have the type of clustering listed as the title for that column. I think that to do that you would have to break up how you are creating the plots and `purrr::map` through the `cluster_types` and then combine the two columns which might get complicated. The font size also is still a little small, but I know you need to fit them all in so I don't know what else we can do about that.

Okay I believe I've fixed everything you mentioned here with exception of the UMAP font sizes (still no luck getting the font bigger there without clipping some of them off). I also made sizes a bit more dynamic (as to make sure they work if only one cluster type is provided), let me know what you think now @allyhawkins and @sjspielman!

[library01_miQC_clustering_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/9777920/library01_miQC_clustering_report.html.zip)

allyhawkins (commented):
These changes look good to me! My only additional comment is about the description for the UMAPs which includes printing out the cluster types included as a comma separated list, like they were entered in the params section, which doesn't look the greatest or make much sense. I think it would be nice to be able to print a bulleted list with a statement "Here clustering was performed using the following clustering types:" I think if you followed that statement with a new chunk that looked like this it should work: (you might need to put `results='asis'` in the chunk label)

```{r results='asis'}
cat(paste0("-", params$cluster_type), sep = "\n")
```

sjspielman (commented):
Looks good! However, I'm noticing something a little unfortunate that maybe we are only seeing for the first time...In the ARI plots, the dodging summary pointranges are correctly offset so they aren't on top of each other. But, at the _ends_ of the lines, this means one pointrange geom is just not on top of anything and it can look ambiguous as to which line it belongs to.
A quick fix here could be coloring the point range, but to make sure we can read everything clearly, it might look nice to use `shape=21` and specify the *fill* aesthetic maps to cluster type rather than color.

This may be something we want to fix elsewhere, too.

cbethell:
Please let me know if I implemented what you both imagined in your latest comments above, @allyhawkins and @sjspielman (especially your comment Stephanie).
The latest updates can be seen in the html here:
[library01_miQC_clustering_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/9780497/library01_miQC_clustering_report.html.zip)

optional-clustering-analysis/clustering-report-template.Rmd
```diff
@@ -141,13 +149,22 @@ summary_stability_stats_df <- summary_stability_stats_df %>%
 ## UMAP plot results for `r params$library`
 
 Below is a set of UMAP plots, where each panel corresponds to clustering performed with a given clustering parameter and each cell is colored based on the cell's cluster assignment.
-Here clustering was performed using `r params$cluster_type` clustering.
+Here clustering was performed using the following clustering types:
+
+```{r results='asis'}
+cat(paste("-", cluster_types), sep = "\n")
+```
```
allyhawkins, line 156: This looks great! One more minor change, can you take this sentence and this chunk and move it to be the last sentence in the intro after line 27?

allyhawkins (approved):
This looks good! I had one minor comment and then can you change the color of the median line in the last summary plot to be black if you are going to color them by cluster type? That way we can see the line? Other than that everything looks good and no need for me to take another look.

sjspielman (approved):
🎉 !

PR #230, opened 2022-10-19
Add GOI analysis calculations script

### LINKED ISSUE #172: Move mapping gene identifiers from separate script to function run within GOI calculations script

Currently in the GOI module we have a separate script that is used for mapping gene identifiers and is an optional step that is run in the bash script. Since we are moving to have this module as a Snakefile that renders the GOI report, and snakemake doesn't like optional steps, we should take what is currently the [`map-goi-list.R` script](https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/main/utils/map-goi-list.R) and turn it into a function that is optionally run inside the [`goi-report-template.Rmd`](https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/main/optional-goi-analysis/goi-report-template.Rmd) upon rendering. 

After moving the current mapping script to be a function, we would source it into the beginning of the template notebook. This would require adjusting the parameters of the template notebook to include an option on whether to include mapping or not as well as input and output identifier types. If that option is specified as true, then mapping would occur prior to plotting for the rest of the notebook. 

The current script should be able to be adjusted into a function and I don't believe any of the arguments should have to change. We will just want it to be a function so we can source it in. We will need to add mapping parameters to the template notebook, and I believe that would mean replacing some of the parameters that are currently there for `gene_symbol_column` and `ensembl_id_column`. We probably want to have an `input_gene_list` parameter instead with a `mapping` option and then options for the input gene list type and output gene list type that are only used if mapping is set to true. 

Alternatively we could keep things as is and then have two separate steps, one for mapping and one for rendering the notebook in the Snakefile, but I don't think snakemake likes optional outputs so this would be less flexible with optional mapping steps. 

cbethell:
**Issue Addressed**
Closes #172 

This PR initiations a `goi-calculations.R` script in the optional-goi-analysis module, and modifies the relevant functions script appropriately.

As noted on issue #172, this PR moves the optional gene identifier mapping from a separate R script to the first step of the GOI analysis module which is the `goi-calculations.R` script.

Within the calculations script is also the preparation of the heatmap matrix and annotation, thus removing the need for any calculations to be done in the goi template notebook.

I have also added a short example GOI list to `example-data/goi-lists` for testing purposes (suggestions for optimization here are welcome)!

There is some added logic here as well, to account for the need for Ensembl ids to filter the matrix and run the `colnames_to_gene_symbols()` function in the script, please let me know if you have any suggestions here we well. See below for reference to the code being referred to here:

Screen Shot 2022-10-20 at 2 16 11 PM

- Does this PR seem to adequately address #172? 
- Are there any suggestions for refactoring?

Here is also a screenshot of the output structure currently within one sample's results folder:

Screen Shot 2022-10-20 at 2 19 01 PM

**Checklist**


optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+    type = "character",
+    default = NULL,
+    help = "path to RDS file containing SCE object with log-normalized counts 
+    and PCA embeddings",
+  ),
+  optparse::make_option(
+    c("-l", "--library_id"),
+    type = "character",
+    default = NULL,
+    help = "name of library being analyzed",
```
allyhawkins, line 41:
```suggestion
    help = "Unique ID corresponding to library being analyzed",
```
Just want to make sure we specify that each library ID should be unique.

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+  ),
+  optparse::make_option(
+    c("-s", "--seed"),
+    type = "integer",
+    default = 2021,
+    help = "seed integer",
```
allyhawkins, line 47:
```suggestion
    help = "Random seed to set",
```

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+    metavar = "integer"
+  ),
+  optparse::make_option(
+    c("-g", "--input_goi_list"),
+    type = "character",
+    default = NULL,
+    help = "file path to the unmapped input genes of interest list"
```
allyhawkins, line 54:
```suggestion
    help = "file path to the input genes of interest list"
```
Doesn't necessarily need to be unmapped

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+    default = NULL,
+    help = "type of gene identifiers to be mapped to and returned"
+  ),
+  optparse::make_option(
+    c("-c", "--identifier_column_name"),
+    type = "character",
+    default = NULL,
+    help = "the name of the column in the input file that contains the gene
+            identifiers to be converted from"
+  ),
```
allyhawkins, line 82: I think we might actually be able to remove this and specify in the docs what the column name should be? We could use a generic `gene_id` for the column name for the input rather than having people set their own column name.

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+  optparse::make_option(
+    c("--gene_set_column_name"),
+    type = "character",
+    default = NULL,
+    help = "the name of the column in the input file that contains the gene set
+            information"
+  ),
```
allyhawkins, line 89: This is an additional grouping right? Like if a gene belongs to a group of genes? We should maybe expand our example file to include a gene set column. Again, I would also set a required name for this column like `gene_set` and make that column optional.

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+            mapped, 'Homo sapiens' or 'Mus musculus' for examples"
+  ),
+  optparse::make_option(
+    c("-m", "--multi_mappings"),
+    type = "character",
+    default = NULL,
+    help = "how should instances of multiple gene identifier mappings be handled
+            - may specify 'list' to return all the mappings or 'first' to keep
+            only the first mapping"
+  ),
```
allyhawkins, line 104: I think here you want to set a default of what we would recommend, which is probably `first`. Also if someone were to use `list` how would that be incorporated into the downstream plots? We probably want to always include a column with the first mapping and then if someone uses `list` then they would receive an additional column with secondary mappings, but the plots will only include the first mapping.

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+  library(AnnotationDbi)
+  library(org.Hs.eg.db)
+  library(ComplexHeatmap)
+  library(SingleCellExperiment)
+})
+
+#### Read in data --------------------------------------------------------------
+
+goi_list <- data.table::fread(opt$input_goi_list, stringsAsFactors = FALSE)
+sce <- read_rds(file.path(opt$sce))
```
allyhawkins, line 169: This is where you would want to add a check to make sure that the SCE object is in the correct format, e.g. contains normalized data

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+  # Save mapped object to file
+  write_tsv(goi_list, file.path(
+    opt$output_directory,
+    paste0(opt$library_id, "_mapped_genes.tsv")
+  ))
+}
+
+#### Prepare data for plotting -------------------------------------------------
+
+identifier_column_name_sym <- rlang::sym(opt$identifier_column_name)
```
allyhawkins, line 228: You don't actually use this anywhere else? I think you could delete it?

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+
+#### Read in data --------------------------------------------------------------
+
+goi_list <- data.table::fread(opt$input_goi_list, stringsAsFactors = FALSE)
+sce <- read_rds(file.path(opt$sce))
+
+#### Perform mapping -----------------------------------------------------------
+
+if (opt$perform_mapping == TRUE) {
+  # turn column name into symbol for pulling the column info out of data frame
```
allyhawkins, line 174: Before performing any mapping you probably want a check that provided key types are present in the annotations.

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+    help = "type of gene identifiers provided in the list -- 'SYMBOL' or
+           'ENSEMBL' for example, see `keytypes()` in the organism's annotation
+            package for more options"
+  ),
+  optparse::make_option(
+    c("--output_identifiers"),
+    type = "character",
+    default = NULL,
+    help = "type of gene identifiers to be mapped to and returned"
+  ),
```
allyhawkins, line 75:
```suggestion
  optparse::make_option(
    c("--sce_rownames_identifier"),
    type = "character",
    default ="ENSEMBL",
    help = "gene identifier type associated with rownames of provided SCE object"
  ),
```
In thinking about how we want to frame this, I think we want to refer to this as the keytype associated with the rownames in the SCE object. The goal of this is to get the input genes in the same format as those row names, which by default in our pipeline will be ensembl gene IDs. You would need to propagate this variable name change throughout.

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+    dplyr::distinct() %>%
+    # join the remaining columns
+    dplyr::left_join(goi_list, by = tolower(opt$identifier_column_name))
+  
+  # Save mapped object to file
+  write_tsv(goi_list, file.path(
+    opt$output_directory,
+    paste0(opt$library_id, "_mapped_genes.tsv")
+  ))
+}
```
allyhawkins, line 224:
```suggestion
  ))
  
  # define mapped goi associated with rownmaes of SCE object 
  goi_rownames_column <- tolower(opt$sce_rownames_identifier)
  goi_rownames <- goi_list %>%
    dplyr::pull(goi_rownames_column)
    
} else {

  # if no mapping is performed then set goi to input id column 
  goi_rownames <- goi_list %>%
    dplyr::pull(gene_id)
  }
```

Here I would set the variables for grabbing the genes from the SCE object based on if mapping was performed or not. I also used my suggestion from earlier on setting the input ID column to always be `gene_id` and removing that argument entirely.

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,301 @@
...
+# we need to filter using the column with Ensembl ids
+if(!is.null(opt$output_identifiers) && (opt$output_identifiers == "ENSEMBL")) {
+  ensembl_id_column <- tolower(opt$output_identifiers)
+} else {
+  ensembl_id_column <- opt$identifier_column_name
+}
+
+# filter counts matrix to only data associated with the provided genes of
+# interest
+normalized_sce_logcounts_matrix <- normalized_sce_logcounts_matrix[,colnames(normalized_sce_logcounts_matrix) %in% goi_list[[ensembl_id_column]]]
```
allyhawkins:
```suggestion
normalized_sce_logcounts_matrix <- normalized_sce_logcounts_matrix[,colnames(normalized_sce_logcounts_matrix) %in% goi_rownames]
```
simplifying a little with my previous suggestion

allyhawkins (commented):
Thanks for getting this started! I have some comments throughout to help narrow down the arguments needed and refactor a little/ add in some checks. I probably have more comments on how to clean up preparation of the heatmap, but want to get these to you right now and then we can work through that part.

cbethell:
Thank you for the review @allyhawkins, it has been extremely helpful!

I was able to implement your refactoring suggestions, add the example `gene_set` column to the goi list added here, and include the necessary checks. 

The only comments I want to shed light on at this time are:
- > I think here you want to set a default of what we would recommend, which is probably first. Also if someone were to use list how would that be incorporated into the downstream plots? We probably want to always include a column with the first mapping and then if someone uses list then they would receive an additional column with secondary mappings, but the plots will only include the first mapping.
https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/230#discussion_r1004640593

As we discussed yesterday, I agree with you here but have not pinned down how to ideally tackle this as yet (with the lack of example multi-mapping data). Let me know if you have anything readily in mind, otherwise I will dig further into this. Also let me know if it would be easier to you to pair on this!

- > This is where you would want to add a check to make sure that the SCE object is in the correct format, e.g. contains normalized data
https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/230#discussion_r1004642646

I added a check that `logcounts(sce)` is not NULL but please let me know if you had more in mind regarding what should be checked for at this point!

Overall let me know what you think of things now!

cbethell:
Committed the further refactoring changes we discussed in Science Pair It @allyhawkins, _now_ I believe this is ready for another look!

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,291 @@
...
+# Command line usage:
+
+# Rscript --vanilla goi-calculations.R \
+#   --sce "example-results/sample01/library01_miQC_processed_sce.rds" \
+#   --library_id "library01" \
+#   --seed 2021 \
+#   --input_goi_list "example-data/goi-lists/sample01_goi_list.tsv" \
+#   --perform_mapping TRUE \
+#   --sce_rownames_identifier "ENSEMBL" \
+#   --provided_identifier <- "SYMBOL" \
```
allyhawkins, line 15:
```suggestion
#   --provided_identifier "SYMBOL" \
```

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,291 @@
...
+    c("-g", "--input_goi_list"),
+    type = "character",
+    default = NULL,
+    help = "file path to the input genes of interest list"
+  ),
+  optparse::make_option(
+    c( "--perform_mapping"),
+    action = "store_false",
+    help = "specifies whether or not to perform gene identifier mapping; the 
+    default is FALSE as mapping is not required when ENSEMBL ids are provided"
```
allyhawkins, line 59:
```suggestion
    action = "store_true",
    default = FALSE,
    help = "specifies whether or not to perform gene identifier mapping; the 
    default is FALSE as mapping is not required when ENSEMBL ids are provided"
```
Based on the description I think what you mean here is to use `store_true` which will set this to true by using the `--perform_mapping` flag at the command line and then it will be false if the flag is not used.

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,291 @@
...
+## embeddings.
+
+# Command line usage:
+
+# Rscript --vanilla goi-calculations.R \
+#   --sce "example-results/sample01/library01_miQC_processed_sce.rds" \
+#   --library_id "library01" \
+#   --seed 2021 \
+#   --input_goi_list "example-data/goi-lists/sample01_goi_list.tsv" \
+#   --perform_mapping TRUE \
```
allyhawkins, line 13:
```suggestion
#   --perform_mapping \
```
Shouldn't need to specify true/false with `action=store_true`

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,291 @@
...
+    type = "integer",
+    default = 2021,
+    help = "Random seed to set",
+    metavar = "integer"
+  ),
+  optparse::make_option(
+    c("-g", "--input_goi_list"),
+    type = "character",
+    default = NULL,
+    help = "file path to the input genes of interest list"
```
allyhawkins, line 53:
```suggestion
    help = "file path to the input genes of interest list, must contain gene identifiers to plot in a column named `gene_id`."
```

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,291 @@
...
+# Rscript --vanilla goi-calculations.R \
+#   --sce "example-results/sample01/library01_miQC_processed_sce.rds" \
+#   --library_id "library01" \
+#   --seed 2021 \
+#   --input_goi_list "example-data/goi-lists/sample01_goi_list.tsv" \
+#   --perform_mapping TRUE \
+#   --sce_rownames_identifier "ENSEMBL" \
+#   --provided_identifier <- "SYMBOL" \
+#   --organism "Homo sapiens" \
+#   --multi_mappings "list" \
```
allyhawkins, line 17:
```suggestion
#   --multi_mappings "first" \
```

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,291 @@
...
+    help = "how should instances of multiple gene identifier mappings be handled
+            - may specify 'list' to return all the mappings or 'first' to keep
+            only the first mapping"
+  ),
+  optparse::make_option(
+    c("--output_directory"),
+    type = "character",
+    default = NULL,
+    help = "path to output directory that would hold the TSV files with the 
+    clustering stats"
```
allyhawkins, line 94:
```suggestion
    help = "path to output directory"
```

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,291 @@
...
+if(is.null(logcounts(sce))){
+  stop("There is no normalized data in the provided SingleCellExperiment object.
+       You may want to run the core workflow analysis before re-running this GOI analysis.")
+}
+
+# Check for dimensionality reduction results
+if(!(any(c("PCA", "UMAP") %in% reducedDimNames(sce)))){
+  stop("There are no PCA or UMAP results in the provided SingleCellExperiment object.
+       You may want to run the core workflow analysis before re-running this GOI analysis.")
+}
```
allyhawkins, line 165: One other check I thought of we are missing is that `gene_id` is a column name in the input GOI list.

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,291 @@
...
+} else {
+  # if no mapping is performed then set goi to input id column
+  goi_rownames <- goi_list %>%
+    dplyr::pull(gene_id)
+}
+
+#### Prepare data for plotting -------------------------------------------------
+
+# get logcounts from normalized sce object
+normalized_sce_logcounts_matrix <- as.matrix(t(logcounts(sce)))
```
allyhawkins, line 235:
```suggestion
normalized_logcounts_matrix <- as.matrix(t(logcounts(sce)))
```
I think throughout you could remove the `_sce_` from the variable names since these are now matrices and not SCE objects.

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,291 @@
...
+  } 
+  column_annotation <- NULL
+}
+
+# Save matrix object to file
+write_rds(normalized_sce_zscores_matrix,
+          file.path(
+            opt$output_directory,
+            paste0(opt$library_id, "_normalized_sce_zscores.rds")
+          ))
```
allyhawkins, line 285: This is a matrix correct? We could save this as a compressed `.mtx` file instead of an Rds file?

allyhawkins (commented):
Just a few more minor comments, but the logic of everything looks good now!

cbethell:
Thanks @allyhawkins, I've implemented your additional suggestions with exception of saving the matrix as a `.mtx` file. I tried to do this using the `Matrix::writeMM()` function but received the following error after confirming the package is indeed installed, any ideas?

Screen Shot 2022-11-02 at 5 15 51 PM

allyhawkins:
> Thanks @allyhawkins, I've implemented your additional suggestions with exception of saving the matrix as a .mtx file. I tried to do this using the Matrix::writeMM() function but received the following error after confirming the package is indeed installed, any ideas?

This is because it looks like the scaled logcounts matrix that you're trying to write out is actually a dense matrix and not a sparse matrix, which is what `writeMM` requires. I would first convert to a sparse matrix and then write it out.

cbethell:
> This is because it looks like the scaled logcounts matrix that you're trying to write out is actually a dense matrix and not a sparse matrix, which is what `writeMM` requires. I would first convert to a sparse matrix and then write it out.

Ah yes, that worked and makes sense why! Thank you @allyhawkins!

allyhawkins (approved):
LGTM 🚀

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,303 @@
...
+    tidyr::unnest(cols = tolower(opt$sce_rownames_identifier)) %>%
+    # grab only the unique rows
+    dplyr::distinct() %>%
+    dplyr::left_join(goi_list, by = "gene_id")
+  
+  # Save mapped object to file
+  write_tsv(goi_list, file.path(
+    opt$output_directory,
+    paste0(opt$library_id, "_mapped_genes.tsv")
+  ))
```
allyhawkins, line 228:
```suggestion
```

As I started to review #232, I realized that we are always taking the mapped goi list as input, but if mapping is false that file won't exist. So what we need to do is move writing out the file to be outside of the if statement and write out a new goi list that gets stored in the goi stats directory regardless. I think we could still name this as with the `_mapped_genes.tsv` extension, but it will only contain the additional mapped columns if mapping was performed, otherwise it will be the same as the input goi list.

optional-goi-analysis/goi-calculations.R
```diff
@@ -0,0 +1,303 @@
...
+  goi_rownames_column <- tolower(opt$sce_rownames_identifier)
+  goi_rownames <- goi_list %>%
+    dplyr::pull(goi_rownames_column)
+  
+} else {
+  # if no mapping is performed then set goi to input id column
+  goi_rownames <- goi_list %>%
+    dplyr::pull(gene_id)
+}
+
```
allyhawkins, line 240:
```suggestion
 # Save goi list to file to be used as input for template 
  write_tsv(goi_list, file.path(
    opt$output_directory,
    paste0(opt$library_id, "_mapped_genes.tsv")
  ))
```

allyhawkins (commented):
Sorry one more change I found we need.

cbethell:
I have fixed the way we save the output mapped file per https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/230#discussion_r1014215860 and also added the `sce_rownames_identifier` column here per https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/232#discussion_r1014238849. I _believe_ this should be ready now!

allyhawkins:
Took a quick glance and the changes look correct! Should be good to go to merge.

PR #248, opened 2022-11-01
Move independent installation instructions to separate doc

### LINKED ISSUE #235: Move independent installation instructions to separate doc

Right now the presence of two types of installation in the main readme can be confusing, we should move those instructions for the independent installation to avoid confusion. 

Moving everything under the `Independent installation` heading to a new doc.

cbethell:
**Issue Addressed**
Closes #235

This PR moves the independent installation instructions to a separate document named `independent-installation-instructions.md`.

This PR moves the `Independent installation` section from the main `README.md` file related to the core worflow, to a separate document named `independent-installation-instructions.md` per issue #235.

The main `README.md` file was also updated to point to this separate document if a user desires to implement independent installation.

- Does this PR seem to adequately address issue #235? Are there any additional details you can think of that may be beneficial here?

**Checklist**


allyhawkins (commented):
Thanks for doing this. A few things before I approve. Can you change this to branch off `development` and then we can do a release with the docs changes? 

And I think we might want to consider starting a `docs` folder where any extra documentation might hang out. So this doc, the processing information doc, and then any other things we add can go there. That way we don't clutter the root directory.

cbethell:
> And I think we might want to consider starting a `docs` folder where any extra documentation might hang out. So this doc, the processing information doc, and then any other things we add can go there. That way we don't clutter the root directory.

Good point @allyhawkins! I've moved the processing info doc and the independent installation instructions doc into a new folder named `additional-docs`. Let me know what you think now!

allyhawkins (approved):
LGTM

PR #252, opened 2022-11-10
Add more details on the mitochondrial gene list

### LINKED ISSUE #236: Include instructions on which mitochondrial gene list to use with data downloaded from scpca-nf

In order to run the core workflow users will need to either provide a mitochondrial gene list or use the gene list that we have provided. We briefly mention this in the `project-specific parameters` section, but don't actually expand on how to obtain the list outside of the one we provide. If users are downloading data from `scpca-nf` we have a mitochondrial gene list for them to use, but depending on the version of the index they were aligned to that could dictate which gene list should be used. 

We should include instructions on how to find out which version of the index, and ultimately which reference mito list, should be used for datasets downloaded from `scpca-nf`. 

We probably want to include a note in the `project-specific parameters` section on how to find out what mito list to use. This should involve instructions to look at the assembly column of the metadata file associated with that library.

cbethell:
**Issue Addressed**
Closes #236

We do not provide much information around the mitochondrial gene list provided and how to obtain one if using other data from `scpca-nf`.

This PR updates the `README.md` file with further details regarding the mitochondrial gene list that is required to run the core workflow and how to obtain the mitochondrial gene list if using other data from `scpca-nf`.

- Please let me know if I am missing any important information here or misconstruing any of the added info.
- Do you have suggestions for any helpful resources to link to at this point?

**Checklist**


README.md
```diff
@@ -183,7 +183,7 @@ These include the following parameters:
...
 | `project_metadata` | relative path to your specific project metadata TSV file |
-| `mito_file` | full path to a file containing a list of mitochondrial genes specific to the genome or transcriptome version used for alignment. By default, the workflow will use the mitochondrial gene list obtained from Ensembl version 104 which can be found in the `reference-files` directory. |
+| `mito_file` | full path to a file containing a list of mitochondrial genes specific to the genome or transcriptome version used for alignment. By default, the workflow will use the mitochondrial gene list obtained from Ensembl version 104 which can be found in the `reference-files` directory. **Note** that the provided mithochondrial gene list is compatible with the example datasets we processed and provided using [`scpca-nf`](https://github.com/AlexsLemonade/scpca-nf#scpca-nf). You may need to generate a new mitochondrial gene list for other data obtained from `scpca-nf` based on the reference used for alignment. To find reference version associated with other data from `scpca-nf`, you will want to look at the `assembly` column of the metadata file associated with that library. |
```
allyhawkins, line 186:
```suggestion
| `mito_file` | full path to a file containing a list of mitochondrial genes specific to the reference genome or transcriptome version used for alignment. By default, the workflow will use the mitochondrial gene list obtained from Ensembl version 104 of the Human transcriptome which can be found in the [`reference-files` directory](./reference-files). 
**Note**The default mithochondrial gene list is compatible with any libraries aligned to the Ensembl version 104 of the Human transcriptome.
For all datasets downloaded from the [ScPCA Portal](https://scpca.alexslemonade.org/), the Ensembl version used can be found by looking at the `assembly` column of the `metadata.json` file associated with that library.
You should not need to change this parameter unless your dataset has been aligned to a different reference.  |
```
I think we want to be a little more detailed here. Generally people should not need to change this default if they are downloading data from the portal, so the main point is that we need to tell them how to find the assembly version from the portal and then tell them what the default version is. On our side if we add another version of ensembl to the reference index for `scpca-nf` we should be sure to include that here so users don't have to make their own if they are downloading data from `scpca-nf`. However, if they are trying to use other data, then they would have to create their own. We probably want to add one more additional sentence stating that if needing to use your own you will need to just grab all possible mitochondrial genes from the reference transcriptome and create a text file with one gene per line.

allyhawkins (commented):
Thanks for doing this! Just a few comments on how to make this a bit more specific and detailed.

README.md
cbethell: Thanks for the review @allyhawkins! I've implemented your suggestions here and took a lot of this wording out of the table as it was getting a little lengthy for the table. Let me know what you think now!

README.md
```diff
@@ -48,7 +48,7 @@ To run the core downstream analyses workflow on your own sample data, you will n
...
 2. Single-cell gene expression data stored as `SingleCellExperiment` objects stored as RDS files (see more on this in the ["Input data format" section](#input-data-format))
 3. A project metadata tab-separated value (TSV) file containing relevant information about your data necessary for processing (see more on this in the ["Metadata file format" section](#metadata-file-format) and an example of this metadata file [here](https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/main/project-metadata/example-library-metadata.tsv))
-4. A mitochondrial gene list that is compatible with your data (see more on this in the ["Running the workflow" section](#running-the-workflow))
+4. A mitochondrial gene list as a text file with a list of mitochondrial genes found in the reference transcriptome used for alignment. Within this file, each row must contain a unique gene identifier corresponding to a mitochondrial gene found in the reference genome used for alignment (see more on this in the ["Running the workflow" section](#running-the-workflow)).
```
allyhawkins, line 51:
```suggestion
4. A mitochondrial gene list as a text file with a list of mitochondrial genes found in the reference transcriptome used for alignment. 
Within this file, each row must contain a unique gene identifier corresponding to a mitochondrial gene found in the reference genome used for alignment (see more on this in the ["Running the workflow" section](#running-the-workflow)).
```
One sentence per line

allyhawkins (approved):
This looks good! Just one minor formatting change.

Issue #250, opened 2022-11-09
Move example metadata file 


cbethell:
_Please label your issue as either `enhancement`, `visualization`, `bug`, or `documentation`._

**Please provide some background on the proposed additions or changes.**

The example metadata file was not easily found during at least one of our pilot eval sessions. We may want to try moving the example project metadata file from `project_metadata` to the `example_data` folder of the repo.

**What are the changes that you are proposing?**

Move the `example-library-metadata.tsv` file from the `project_metadata` to the `example_metadata` folder and update documentation where applicable.

We may also want to hide the `project_specific_files` folder while addressing this issue as users may be confused on why it is there and what is in it.

cbethell:
Addressed by merged PR #255

PR #255, opened 2022-11-11
Move filtering method to config file

### LINKED ISSUE #243: Move filtering method from metadata to config file

The filtering method is currently required to be listed in the metadata file for each library being processed. This used to be required when we didn't account for failure of `miQC` so each library had to be specified individually in case one failed. However, it doesn't quite make sense now given the addition of all the other params in the config file for it to be in the metadata file and not be in the config file. 

The filtering method should be moved out of the metadata file and into the config file. This will also require changes in the snakemake workflow and how the filtering method is read in. While doing this we should consider changing the workflow from labeling each file with the filtering method since it is included in the report. 

This will also require changes to the docs since we currently list the filtering method as being required in the metadata of the workflow.

cbethell:
**Issue Addressed**
Closes #243

The purpose of this PR is to move the filtering method from the metadata to the config file as it is no longer necessary in the metadata file now that we have a check that uses manual filtering if miQC fails.

This PR does the following:

- moves the filtering method "miQC" from the metadata file to the `config/config.yaml` file
- updates the `README.md` file to reflect this change
- updates the core Snakefile and the clustering Snakefile to reflect this change (I have tested that both run successfully with the new changes in this PR)
- moves the example project metadata file into `example-data` per issue #250

- Does this PR seem to satisfy what is outlined on issue #243?

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,6 @@ The file should contain the following columns:
 
 - `sample_id`, unique ID for each piece of tissue or sample that cells were obtained from,  all libraries that were sampled from the same piece of tissue should have the same `sample_id`.
 - `library_id`, unique ID used for each set of cells that has been prepped and sequenced separately.
-- `filtering_method`, the specified filtering method which can be one of "manual" or "miQC". For more information on choosing a filtering method, see [Filtering low quality cells](./processing-information.md#filtering-low-quality-cells) in the [processing information documentation](./processing-information.md).
 - `filepath`, the full path to the RDS file containing the pre-processed `SingleCellExperiment` object.
 Each library ID should have a unique `filepath`.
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -229,6 +228,7 @@ Below are the parameters required to run either of the filtering methods.
 | Parameter        | Description | Default value |
 |------------------|-------------|---------------|
 | `seed` | an integer to be used to set a seed for reproducibility when running the workflow | 2021 |
+| `filtering_method` | `filtering_method`, the specified filtering method which can be one of "miQC" or "manual". For more information on choosing a filtering method, see [Filtering low quality cells](./processing-information.md#filtering-low-quality-cells) in the [processing information documentation](./processing-information.md) | "miQC" |
```
allyhawkins, line 231: I think we are also going to need to edit the processing information doc in the filtering section. There's a line that says the filtering method is specified in the metadata file which is no longer the case.

allyhawkins: Along that same note, we need to update the expected output structure here too to remove filtering method from the file name.

allyhawkins (commented):
The actual change to move the parameter looks good. I think what's missing is some docs changes.

cbethell:
Thanks for the review @allyhawkins! I implemented those docs changes you pointed out (and double-checked to see if I am missing anymore -- I think we got them all but please let me know if I am wrong)!

PR #261, opened 2022-11-16
Document core analysis template report

### LINKED ISSUE #239: Document the core workflow template

The template produced by the core workflow is lacking a lot of documentation and explanation of the plots. The template should be updated to include a description of the expected plots. 

In particular a description of the filtering plot and UMAP is needed. We will need to think about how we want to address describing the filtering plot because different methods of filtering can be used.

cbethell:
**Issue Addressed**
Closes #239

The core analysis report should be properly documented to give users an idea of how to interpret the plots in the report.

This PR adds details about the filtering plots, explaining the plots will be different whether `miQC` or `manual` filtering is implemented.

This PR also cleans up the report a bit by removing the numbering before each of the headers and hiding the set up chunks as there are no plots or useful information to be displayed there.

- Does the description above the filtering plots make sense to you? Is it easy to interpret?
- The description above the UMAP plots seems suffice as is (general enough in case users modify the clustering parameters), but please let me know if you feel additional information should be added there.

Here is an example of the report when miQC filtering is implemented:
[library01_core_analysis_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/10025498/library01_core_analysis_report.html.zip)

Here is an example of the report when manual filtering is implemented:
[library01_core_analysis_report_manual_example.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/10025500/library01_core_analysis_report_manual_example.html.zip)

**Checklist**

### DIFF
--- a/core-analysis/core-analysis-report-template.Rmd
+++ b/core-analysis/core-analysis-report-template.Rmd
@@ -17,11 +17,11 @@ output:
     toc: true
     toc_depth: 2
     toc_float: true
-    number_sections: true
+    number_sections: false
     code_folding: hide
 ---
 
-## Set Up
+This notebook will plot the filtering and clustering results from the core analysis workflow.
 
 ```{r, setup, include=FALSE}
 knitr::opts_chunk$set(
@@ -46,7 +46,7 @@ source(file.path(project_root, "utils", "setup-functions.R"))
 source(file.path(project_root, "utils", "filtering-functions.R"))
 ```
 
-```{r}
+```{r include=FALSE}
 # Load project
 setup_renv(project_filepath = project_root)
 
@@ -55,7 +55,7 @@ library(ggplot2)
 theme_set(theme_bw())
 ```
 
-```{r}
+```{r include=FALSE}
 # Load libraries
 library(readr)
 library(SingleCellExperiment)
@@ -65,7 +65,7 @@ library(ggpubr)
 library(scater)
 ```
 
-```{r}
+```{r include=FALSE}
 # Read in SCE objects
 pre_processed_sce <- read_rds(file.path(params$pre_processed_sce))
 
@@ -134,6 +134,16 @@ knitr::kable(sample_information, align = 'r') %>%
 ```
 
 ### Filtering plots
+
+The filtering plots below represent the cutoffs used to filter and remove low-quality cells.
+When `miQC` filtering is implemented, two plots are displayed.
+The first plot shows the cutoff used for the maximum probability of a cell being compromised.
+The second plot indicates which cells have been removed using binary values.
+When `manual` filtering is implemented, one plot is displayed.
+The plot in this case represents the manual filtering cutoffs for maximum percent mitochondrial reads per cell and the minimum number of genes detected per cell.
+
+Here, cells were `r processed_meta$filtering`.
+
 
 ```{r}
 # Read in mito genes
--- a/core-analysis/core-analysis-report-template.Rmd
+++ b/core-analysis/core-analysis-report-template.Rmd
@@ -144,6 +144,8 @@ Cells labeled with `TRUE` are kept for downstream analysis, while all other cell
 When `manual` filtering is implemented, one plot is displayed.
 The plot in this case represents the manual filtering cutoffs for the minimum number of genes detected per cell and the minimum unique molecular identifiers (UMI) per cell.
 The lines on the plot indicate the cutoffs, where cells with number of genes expressed and unique molecular identifiers (UMI) that are both above the minimum cutoffs are kept for downstream analysis.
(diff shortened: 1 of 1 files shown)


core-analysis/core-analysis-report-template.Rmd
```diff
@@ -135,6 +135,16 @@ knitr::kable(sample_information, align = 'r') %>%
 
 ### Filtering plots
 
+The filtering plots below represent the cutoffs used to filter and remove low-quality cells.
+When `miQC` filtering is implemented, two plots are displayed.
+The first plot shows the cutoff used for the maximum probability of a cell being compromised.
```
allyhawkins, line 140:
```suggestion
`miQC` relies on fitting a mixture model using the number of genes expressed by a cell and the percentage of mitochondrial reads. 
The first plot will show the model fit by `miQC` with two model fit lines and each cell being colored by the probability of that cell being compromised. 
```
The first plot is actually showing the model produced by miQC and the cells are colored by the probability of being compromised. The second plot is where the cutoff is used to determine if cells are kept or removed.

core-analysis/core-analysis-report-template.Rmd
```diff
@@ -135,6 +135,16 @@ knitr::kable(sample_information, align = 'r') %>%
...
+The second plot indicates which cells have been removed using binary values.
```
allyhawkins, line 141:
```suggestion
The second plot indicates which cells have been removed based on the cutoff used for a cell's probability of being compromised.
Cells labeled with `TRUE` are kept for downstream analysis, while all other cells were removed. 
```
We should make sure to state exactly what the filtering thresholds are and then where they come from.

core-analysis/core-analysis-report-template.Rmd
```diff
@@ -135,6 +135,16 @@ knitr::kable(sample_information, align = 'r') %>%
...
+When `manual` filtering is implemented, one plot is displayed.
+The plot in this case represents the manual filtering cutoffs for maximum percent mitochondrial reads per cell and the minimum number of genes detected per cell.
```
allyhawkins, line 143: I would also say that the lines indicate the cutoffs used and that cells in the upper left quadrant and those with mito content below the max cutoff are kept for further analysis.

allyhawkins (commented):
Thanks for doing this, it looks so much better already! I just had some comments on clarifying the filtering plots.

cbethell:
Great points @allyhawkins! I've made the appropriate changes and dug into the plotting functions to ensure that we are indeed giving users the right information to interpret the plots.

Particularly referencing your above comment:
> I would also say that the lines indicate the cutoffs used and that cells in the upper left quadrant and those with mito content below the max cutoff are kept for further analysis.

I double-checked the `plot_manual_filtering()` function in the `filtering-functions.R` script and it looks like I misinterpreted things the first time around, the cutoffs are specific to detected genes and UMI (see below).

Screen Shot 2022-11-18 at 12 31 19 AM

I updated the documentation in the template to reflect this, but please let me know if you have suggestions for alternate wording (or any suggestions otherwise)!

core-analysis/core-analysis-report-template.Rmd
```diff
@@ -135,6 +135,19 @@ knitr::kable(sample_information, align = 'r') %>%
...
 
+The filtering plots below represent the cutoffs used to filter and remove low-quality cells.
+When `miQC` filtering is implemented, two plots are displayed.
+`miQC` relies on fitting a mixture model using the number of genes expressed by a cell and the percentage of mitochondrial reads. 
+The first plot will show the model fit by `miQC` with two model fit lines and each cell being colored by the probability of that cell being compromised.
+The second plot indicates which cells have been removed based on the cutoff used for a cell's probability of being compromised.
+Cells labeled with `TRUE` are kept for downstream analysis, while all other cells were removed.
+When `manual` filtering is implemented, one plot is displayed.
+The plot in this case represents the manual filtering cutoffs for the minimum number of genes detected per cell and the minimum unique molecular identifiers (UMI) per cell.
+The lines on the plot indicate the cutoffs, where cells with number of genes expressed and unique molecular identifiers (UMI) that are both above the minimum cutoffs are kept for downstream analysis.
```
allyhawkins, line 146:
```suggestion
The lines on the plot indicate the cutoffs, where cells with number of genes expressed and unique molecular identifiers (UMI) that are both above the minimum cutoffs are kept for downstream analysis.
Additionally, cells are colored by percent mitochondrial reads. 
Cells with percent mitochondrial reads above the provided cutoff are removed prior to downstream analysis.
```
We also remove cells with high mito content and should state that here.

allyhawkins (approved):
The additions you made look good regarding the cutoff. I just had one more suggestion on including the mito cutoff, but I don't need to see them again.

PR #271, opened 2022-12-14
Fix how we handle annotation when gene set is NULL

### LINKED ISSUE #267: Fix how we handle column annotation when `gene_set` is NULL

Currently, we write out a NULL column annotation file when a `gene_set` column is not provided (which is not a great practice). So we will want to modify things to simply not write out a column annotation file when `gene_set` is not provided and account for this in the template notebook as well.

Modify the `goi-calculations.R` script to write out a column annotation file _only if_ a `gene_set` column is provided.
Then we will want to modify `goi-report-template.Rmd` to only read in a column annotation file if the `gene_set` column is present with the associated GOI list.

This will require some added logic related to if a `gene_set` column is present or not.

cbethell:
**Issue Addressed**
Closes #267 

This PR fixes how we handle the heatmap column annotation when `gene_set` is NULL. Instead of saving an empty column annotation file at the end of the GOI calculations script, this PR adds in a few checks to only save the file if gene set and therefore a column annotation exists.

The `goi-calculations.R` script was modified to only save a column annotation file if column annotation exists as a result of the gene set logic.
The `goi-report-template.Rmd` notebook was modified to check if a `gene_set` column exists in the mapped goi list and only reads in a column annotation file if it does.
Noting that if it doesn't, heatmap_annotation is set to NULL in the notebook as to have something to provide to the heatmap plotting function. I did this to avoid having an if statement where there heatmap plotting function is copied and pasted without the annotation argument, but please let me know if you feel this would be better suited here!

Does this PR seem to address #267? 
Do you have any ideas for further refactoring?

**Checklist**

### DIFF
--- a/optional-goi-analysis/goi-calculations.R
+++ b/optional-goi-analysis/goi-calculations.R
@@ -292,8 +292,7 @@ if(!is.null(goi_list$gene_set)) {
         goi_rownames_column,
         "gene_id"
       )
-  } 
-  column_annotation <- NULL
+  }
 }
 
 # Convert heatmap matrix to a sparse matrix and save to file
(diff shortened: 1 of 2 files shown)


allyhawkins (approved):
This all looks good, thanks for doing this!

Issue #273, opened 2022-12-15
Include core workflow example output files to repo


cbethell:
**Please provide some background on the proposed additions or changes.**

It would be helpful to include the example output results files, from the core analysis workflow, to the repository.

**What are the changes that you are proposing?**

This should involve removing the `example-results` directory from the `.gitignore` file.

**Please describe the proposed solution.**

We will want to remove `example-results` from the gitignore file before the next scheduled external usability eval.

PR #276, opened 2022-12-16
Add example results files to repo/Ignore project-specific files

### LINKED ISSUE #272: Ignore `project-specific-files` directory

We currently have a `project-specific-files` directory in this repo that is not required to run the analysis with the example data and therefore can be confusing to external users.

We should add the `project-specific-files` directory to the `.gitignore` file to handle this issue.

This should be a hot-fix PR to ignore the directory before the next scheduled external usability eval.

cbethell:
**Issue Addressed**
Closes #272 and #273

This PR adds the example results html files from the core workflow to the repo. 

This PR removes the `example-results` folder from the `.gitignore` and adds the `project-specific-files` folder there, which in turn adds/removes the appropriate files from the repo.

I did try to include the .rds files in the example results but got a GitKraken error that the `pre-receive hook declined`. I only got things working by removing the .rds files from the commit (even the .rds file under 25MB). 

Any suggestions on getting the .rds files in here?

**Checklist**


allyhawkins:
Is the purpose to have the example results files on the repo itself? Or have them available on the server for easy access during evals? Because if the reasoning is to have them for evals then we could create a symlink for the test account that links to the files in our shared folder that have been processed. 

If we want them on the repo for other potential users, then maybe we try a zip file? Or we have a zip file that lives in a public S3 bucket and provide a link on the repo for users to download (we have an S3 bucket with example scpca-nf output that we could use for this too I think).

cbethell:
> Is the purpose to have the example results files on the repo itself? Or have them available on the server for easy access during evals? Because if the reasoning is to have them for evals then we could create a symlink for the test account that links to the files in our shared folder that have been processed.
> 
> If we want them on the repo for other potential users, then maybe we try a zip file? Or we have a zip file that lives in a public S3 bucket and provide a link on the repo for users to download (we have an S3 bucket with example scpca-nf output that we could use for this too I think).

The purpose of filing the original issue was to have example results files for easy access during evals, however, I _believe_ it would be helpful to also have them accessible to folks just looking at the repo (which folks sometimes would rather during evals). That said, I do like the idea of the example results living in a public S3 bucket that we can provide the link to, and perhaps we should create a symlink for the RStudio test account to link to the shared folder as well @allyhawkins.

allyhawkins:
> The purpose of filing the original issue was to have example results files for easy access during evals, however, I believe it would be helpful to also have them accessible to folks just looking at the repo (which folks sometimes would rather during evals). That said, I do like the idea of the example results living in a public S3 bucket that we can provide the link to, and perhaps we should create a symlink for the RStudio test account to link to the shared folder as well @allyhawkins.

Okay so what I think we want to do is create a folder with the example results in the shared folder that we already have and link that to the training account for user evals. Then take that same folder, zip it up, and copy to S3 so that we can produce a single link that can be used for downloads. We will then need to include a section in the readme on how to download the example results files. I think we should take care of step 1 first (making the example files available on the test server) and file a separate issue for making a copy of the example results available on the repo. 

For making the example files available on the test server I would create a script that runs the snakemake workflow and then creates the link (similar to [`link-data.sh` in training-modules](https://github.com/AlexsLemonade/training-modules/blob/master/scripts/link-data.sh)). We will inevitably have to do this again as updates are made to the workflow and will want to this for the optional modules so building a script that can run the workflows to create the example output and the links will be helpful. We can then build on that script to sync to S3 in the next PR/issue.

cbethell:
> For making the example files available on the test server I would create a script that runs the snakemake workflow and then creates the link (similar to [link-data.sh in training-modules](https://github.com/AlexsLemonade/training-modules/blob/master/scripts/link-data.sh)). We will inevitably have to do this again as updates are made to the workflow and will want to this for the optional modules so building a script that can run the workflows to create the example output and the links will be helpful. We can then build on that script to sync to S3 in the next PR/issue.

Thanks for the reallyy helpful feedback @allyhawkins! So far I have created a script named `utils/sync-results.sh` to sync the results files from the `shared/data` directory to the local directory. I have tested that this works on the test account but left out a step that you mentioned -- running the snakemake workflow. I wanted to clarify this point since it is a step that takes a little while and would essentially be producing the same files that are already in `shared` -- is the idea here to run the latest workflow and have updated files, in which case the syncing from `shared` would not work since the files would already exist from running the workflow?

allyhawkins:
> I have tested that this works on the test account but left out a step that you mentioned -- running the snakemake workflow. I wanted to clarify this point since it is a step that takes a little while and would essentially be producing the same files that are already in shared -- is the idea here to run the latest workflow and have updated files, in which case the syncing from shared would not work since the files would already exist from running the workflow?

Ah yes, you are right that snakemake will complain if the files are already there. I think we just want to make sure we have a system for keeping the files that are being synced up to date and that they are the most recent files. 

I think the script you have looks good, but we need to be sure that it is run when logged into the test account. So I think we will want to have instructions somewhere that state the steps you need to take to make sure example files are up to date: 
1. run the most up to date version of the snakemake workflow(s)
2. run the linking script on the test account 

I think it makes sense to add this to the release checklist? Unless you have other ideas?

cbethell:
I think the release checklist is good place to add those notes! I updated it in the latest commit https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/276/commits/34fe4036d47cbb93f40e3e54e79d74007aea114f, let me know what you think now @allyhawkins!

allyhawkins (commented):
I made a few edits to the checklist you added to clarify some steps, but feel free to change some of those edits if you disagree. 
Working on this made me think we still have room for error. We could leave it as is and then when we work on the S3 stuff think about if we want to make any changes which I think is fine since it does address the point of getting the example files on the test server. 
Or we could adjust the script so that it does two syncs, one from your example-results file (non test account) to the shared directory and then the second from the shared directory specifically to the test account. This means we would add a variable for defining the path to the test account. I don't know if the account is setup to have permissions for us to do that though, so we probably should just leave it as is right now.

cbethell:
Good points @allyhawkins, I've made the more immediate changes to the checklist as you suggested and made a note on the open issue tracking the S3 sync stuff to account for an additional sync in the relevant script https://github.com/AlexsLemonade/scpca-downstream-analyses/issues/277#issuecomment-1377641447.

allyhawkins (approved):
LGTM 🚀

Issue #274, opened 2022-12-15
Emphasize the parts that user will need to replace in example commands


cbethell:
**Please provide some background on the proposed additions or changes.**

We will want to further emphasize the parts of the example commands in the core workflow docs that users will need to replace to successfully run the workflow.

**What are the changes that you are proposing?**

We can change the format of the specific parts of the command that need to replace to something like:

`<REPLACE_WITH_RELATIVE_PATH>`

**Please describe the proposed solution.**

More specifically, the example command under the Quick Start Guide section of the main README would look something like the following:

```
snakemake --cores 2 \
  --use-conda \
  --config results_dir="<REPLACE_WITH_RELATIVE_PATH_TO_RESULTS_DIRECTORY>" \
  project_metadata="<REPLACE_WITH_RELATIVE_PATH_TO_YOUR_PROJECT_METADATA_TSV>"
```

We will want to propagate this format throughout the documentation before the next scheduled external usability eval.

cbethell:
Addressed by #278

Issue #263, opened 2022-11-29
Note minimum snakemake version that is required in the documentation


jaclyn-taroni:
I performed functional testing of setup and running with the example data on my Ubuntu desktop. A snag I ran into, that I was able to identify as a `snakemake` version issue, is that the argument ` --conda-create-envs-only` didn't exist when I ran the `setup_envs.sh` script. We should:

* Identify what version of snakemake `--conda-create-envs-only` first showed up in
* Note this minimum version in this section of the documentation: https://github.com/AlexsLemonade/scpca-downstream-analyses#2-install-snakemake

cbethell:
Addressed by #278.

PR #278, opened 2023-01-06
Update core analysis docs

### LINKED ISSUE #275: Number the preparation steps of the core workflow

It may be helpful to number the steps of the core analysis workflow preparation in the main README to help external users follow along in order.

Number each of the larger headings of the core workflow documentation, as to suggest that steps should be completed in chronological order.

Numbering should start in the sections after the Quick Start Guide, and we will ideally complete this before the next scheduled external usability eval.

cbethell:
**Issue Addressed**
Closes #275, #274, and #263

This PR makes the recommended updates to the main `README.md` file in preparation for further usability evals.

The following changes are made in this PR:
- the main steps of setting up and running the workflow are numbered for easier interpretation (#275)
- emphasis is placed on the parts of the example commands that users would need to replace (#274)
- the minimum Snakemake version is noted as determined by tracking the changelog and the [PR filed](https://github.com/snakemake/snakemake/pull/493) to update Snakemake's documentation (#263)

Do these changes seem to adequately address the issues mentioned/tagged above?

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,8 +63,8 @@ The workflow can directly take as input the `filtered` RDS files downloaded from
 ```
 snakemake --cores 2 \
   --use-conda \
-  --config results_dir="relative path to relevant results directory" \
-  project_metadata="relative path to your-project-metadata.TSV"
+  --config results_dir="<REPLACE_WITH_RELATIVE_PATH_TO_RESULTS_DIRECTORY>" \
+  project_metadata="<REPLACE_WITH_RELATIVE_PATH_TO_YOUR_PROJECT_METADATA_TSV>"
 ```
 
 **Note** that R 4.1 is required for running our pipeline, along with Bioconductor 3.14.
@@ -85,9 +85,9 @@ There are two expected output files thay will be associated with each provided `
 
 See the [expected output section](#expected-output) for more information on these output files.
 
-## How to install the core downstream analyses workflow
+## 1. How to install the core downstream analyses workflow
 
-### 1) Clone the repository
+### a) Clone the repository
 
 First you will want to clone the [`scpca-downstream-analyses` repository](https://github.com/AlexsLemonade/scpca-downstream-analyses) from GitHub.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,8 +63,8 @@ The workflow can directly take as input the `filtered` RDS files downloaded from
 ```
 snakemake --cores 2 \
   --use-conda \
-  --config results_dir="relative path to relevant results directory" \
-  project_metadata="relative path to your-project-metadata.TSV"
+  --config results_dir="<REPLACE_WITH_RELATIVE_PATH_TO_RESULTS_DIRECTORY>" \
+  project_metadata="<REPLACE_WITH_RELATIVE_PATH_TO_YOUR_PROJECT_METADATA_TSV>"
```
allyhawkins, line 67: I actually think I would remove the underscores in between the words. I'm also not crazy about the all caps, but maybe that will draw people's attention to it. 

I also think you want to add a few sentences below that state: "Where `config_results_dir` is the relative path to the results directory where all results from running the workflow will be stored". And then do the same for the project metadata. 
We do something similar in `scpca-nf` - https://github.com/AlexsLemonade/scpca-nf/blob/main/external-instructions.md#configuring-scpca-nf-for-your-environment

README.md
```diff
@@ -100,15 +100,16 @@ More instructions on cloning a GitHub repository can be found [here](https://doc
 
 Once the repository is successfully cloned, a folder named `scpca-downstream-analyses` containing a local copy of the contents of the repository will be created.
 
-### 2) Install Snakemake
+### b) Install Snakemake
 
 The core downstream single-cell analysis pipeline, which includes filtering, normalization, dimensionality reduction, and clustering is implemented using a Snakemake workflow.
 Therefore, you will also need to install Snakemake before running the pipeline.
+Note that the **minimum** version of Snakemake you will need installed to be compatible with conda is version **5.23.0**.
```
allyhawkins, line 107:
```suggestion
Note that the **minimum** version of Snakemake you will need to have installed to be compatible with conda is version **5.23.0**.
```

README.md
```diff
@@ -209,9 +210,9 @@ The below code is an example of running the Snakemake workflow using the project
 ```
 snakemake --cores 2 \
   --use-conda \
-  --config results_dir="relative path to relevant results directory" \
-  project_metadata="relative path to your-project-metadata.TSV" \
-  mito_file="full path to your-mito-file.txt"
+  --config results_dir="<REPLACE_WITH_RELATIVE_PATH_TO_RESULTS_DIRECTORY>" \
```
allyhawkins, line 213: Looking at the fact that we have this multiple places, I wonder if it would actually be better to include a bolded note under the command reminding users that they will need to replace the paths for `results_dir` and `project_metadata`. And then I would make the following change instead: 
```suggestion
  --config results_dir="<RELATIVE PATH TO RESULTS DIRECTORY>" \
```

allyhawkins (commented):
I think rather than changing to the all caps, underscore separated paths in the examples, we should include notes about what the individual arguments represent and then a reminder to replace the paths in bold immediately under the command. I think by explicitly stating it we might get through to more people and we should be sure to explain what each option actually represents.

cbethell:
I've implemented your suggestions to add a line or two explaining the parameters and bolding/emphasizing the need to replace with the relevant filepaths @allyhawkins -- I did keep the CAPS since that was an original suggestion from Deepa but agree that the underscores were much so I removed them. Let me know what you think now!

README.md
```diff
@@ -63,10 +63,14 @@ The workflow can directly take as input the `filtered` RDS files downloaded from
...
   --use-conda \
-  --config results_dir="relative path to relevant results directory" \
-  project_metadata="relative path to your-project-metadata.TSV"
+  --config results_dir="<RELATIVE PATH TO RESULTS DIRECTORY>" \
+  project_metadata="<RELATIVE PATH TO YOUR PROJECT METADATA TSV>"
 ```
 
+Where `results_dir` is the relative path to the results directory where all results from running the workflow will be stored and `project_metadata` is the relative path to the TSV file containing the relevant information about your input files.
+See more information on project metadata in [section 3](#3-metadata-file-format) below.
+**You will want to replace the paths for both `results_dir` and `project_metadata` to successfully run the workflow.**
```
allyhawkins, line 72: I think we want to switch the order of these sentences so the warning comes first.

allyhawkins (approved):
This looks good! Just one sentence order switch, but no need for me to look at it again.

PR #279, opened 2023-01-11
Initiate GOI Snakefile - Calculations Rule

### LINKED ISSUE #229: Initiate snakefile for GOI module with rule to run calculations script

Once we have addressed #172 and have a separate calculations script with optional mapping for the GOI module, we will want to initiate a Snakefile with the first rule being to run the calculations script.

We will want to add a Snakefile to the GOI analysis module and have the first rule run through the GOI calculations script once created as mentioned in https://github.com/alexslemonade/scpca-downstream-analyses/issues/172#issuecomment-1281305712 and similarly to what we did in the clustering module.

As in the clustering module, we will likely want the Snakefile related to the GOI module to live in the root directory while the script and template would be stored in GOI analysis subdirectory.

cbethell:
**Issue Addressed**
Closes #229

This PR initiates the Snakefile that runs the optional GOI analysis module. Only the first rule of the Snakefile is added in this PR, the GOI calculations rule.

- A `goi.snakefile` file is added in this PR and modified to run the `goi-calculations.R` script as the first rule of the Snakefile
- A `goi_config.yaml` file is added to populate the GOI specific parameters in the Snakefile

I did encounter an issue when using the config file to set `organism` = "Homo sapiens". I got an error that `sapiens` cannot be used so it looks like it's breaking up the string when it is passed from the config file to the Snakefile -- this is why I have this filed as a draft PR. I put a quick fix in the calculations script for the purpose of successfully running the Snakefile otherwise but I do not intend on leaving that there. Any suggestions for resolving this?

**Edit:** Solved the above error with a couple of extra `''`

- Does this seem to address #229 

**Checklist**


goi.snakefile
```diff
@@ -0,0 +1,58 @@
...
+               sample = SAMPLES,
+               library = LIBRARY_ID)
+
+rule calculate_goi:
+    input:
+        "{basedir}/{library_id}_processed_sce.rds"
+    output:
+        matrix = "{basedir}/{library_id}_goi_stats/{library_id}_normalized_zscores.mtx",
+        mapped_genes = "{basedir}/{library_id}_goi_stats/{library_id}_mapped_genes.tsv",
+        annotation = "{basedir}/{library_id}_goi_stats/{library_id}_heatmap_annotation.rds",
```
allyhawkins, line 41: You could then remove these lines if you simplify the target rule.

allyhawkins (commented):
This looks pretty good, we just have some extra output files listed that we probably don't need.

PR #285, opened 2023-01-25
Add overwrite option to GOI script

### LINKED ISSUE #282: Add overwrite option to GOI workflow

We may want to add an `--overwrite` option to the current GOI analysis workflow. Reasoning being, if users want to test more than one GOI list, they may want the option to overwrite or not overwrite the current GOI results files.

There should be a check on whether or not `_normalized_zscores.mtx` file exists, if it does, there should be an option to overwrite it or not.

This `--overwrite` option and check for the `_normalized_zscores.mtx` should be added to the `goi_calculations.R` script. 
The default for this option should be stored as `TRUE` in the GOI config file.

Not that I can think of.

cbethell:
**Issue Addressed**
Closes #282

This PR adds the `--overwrite` option to the `optional-goi-analysis/goi-calculations.R` script. This option allows us to account for all three expected output files -- if the `overwrite` option is not provided/is false, then there is a check to see if any of the three expected output files exist and a message to supply the `--overwrite` flag if any of them do exist.

This PR adds the `--overwrite` optparse option, defines the file paths of the expected output files at the beginning of the `goi-calculations.R` script, and adds a check to determine if any of the three output files already exist when `overwrite` is false/not provided.

This PR also updates the GOI config file and GOI `README` to reflect the added `overwrite` option.

- Does this PR seem to appropriately address #282?

**Checklist**

### DIFF
--- a/optional-goi-analysis/README.md
+++ b/optional-goi-analysis/README.md
@@ -117,7 +117,7 @@ The parameters found in the `config/goi_config.yaml` file can be optionally modi
 | `sce_rownames_identifier` | the type of gene identifiers found in the rownames of the SingleCellExperiment object | "ENSEMBL" |
 | `perform_mapping` | a binary value indicating whether or not to perform gene identifier mapping | `TRUE` |
 | `multi_mappings` | how to handle multiple gene identifier mappings when `perform_mapping` is `TRUE` | "list" |
-| `overwrite` | a binary value indicating whether or not to overwrite any existing output files | `TRUE` |
+| `overwrite` | a binary value indicating whether or not to overwrite existing output files | `TRUE` |
 
 
 |[View Genes of Interest Config File](../config/goi_config.yaml)|
--- a/optional-goi-analysis/README.md
+++ b/optional-goi-analysis/README.md
@@ -10,7 +10,11 @@ This directory includes a genes of interest analysis workflow to help users eval
 **Table of Contents**
 
 - [Analysis overview](#analysis-overview)
+- [Running the workflow](#running-the-workflow)
 - [Expected input](#expected-input)
+- [Parameters and config file](#parameters-and-config-file)
+  - [Project-specific parameters](#project-specific-parameters)
+  - [Genes of Interest parameters](#genes-of-interest-parameters)
 - [Expected output](#expected-output)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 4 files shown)


optional-goi-analysis/goi-calculations.R
```diff
@@ -133,11 +139,44 @@ if (!file.exists(opt$sce)){
...
+# Create a list of the expected output files
+output_files <- c(output_mapped_genes, output_matrix, output_annotation)
+
+# Check that any of the output files exist, only if the overwrite flag is false
+if (!opt$overwrite) {
+  if (any(file.exists(output_files))) {
+    existing_file <- output_files[file.exists(output_files)]
+    stop(
+      glue::glue(
+        "{existing_file} file exists, to overwrite please use the --overwrite flag
```
sjspielman, line 172: If there are > 1 existing files, this will print out the same statement a few times. I might have separate printing here for the existing file indication and the `stop()` itself, to avoid overly-repetitive output.

```suggestion
glue::glue("{existing_file} file already exists.")
stop("To re-run analysis and overwrite all existing files please use the --overwrite flag.")
```

optional-goi-analysis/README.md
```diff
@@ -117,6 +117,7 @@ The parameters found in the `config/goi_config.yaml` file can be optionally modi
 | `sce_rownames_identifier` | the type of gene identifiers found in the rownames of the SingleCellExperiment object | "ENSEMBL" |
 | `perform_mapping` | a binary value indicating whether or not to perform gene identifier mapping | `TRUE` |
 | `multi_mappings` | how to handle multiple gene identifier mappings when `perform_mapping` is `TRUE` | "list" |
+| `overwrite` | a binary value indicating whether or not to overwrite any existing output files | `TRUE` |
```
sjspielman, line 120:
```suggestion
| `overwrite` | a binary value indicating whether or not to overwrite existing output files | `TRUE` |
```

optional-goi-analysis/goi-calculations.R
```diff
@@ -133,11 +139,44 @@ if (!file.exists(opt$sce)){
   stop(paste(opt$sce, "does not exist."))
 }
 
-# Check that the outout directory exists
+# Check that the output directory exists
 if (!dir.exists(opt$output_directory)) {
   dir.create(opt$output_directory, recursive = TRUE)
 }
 
+# Define the output file paths
```
sjspielman, line 147: We should add a quick check above here that `opt$library_id` is not `NULL` and print an informative message if it's missing.
Somewhat related (can't formally comment on it here since not a diff), the `library_id` option help should probably be lowercase ("unique" and not "Unique") for consistency with the other option descriptions.

sjspielman (commented):
Looks pretty good! Left you some small comments about phrasing and one more check to add.

optional-goi-analysis/goi-calculations.R
cbethell: I added a `print()` function to get this working -- now it prints out each of the files that already exist and the "To re-run analysis..." statement just once. 👍

cbethell:
Thanks for the feedback @sjspielman! I've implemented those suggestions so let me know what you think of things now!

optional-goi-analysis/goi-calculations.R
```diff
@@ -133,11 +139,45 @@ if (!file.exists(opt$sce)){
...
+)
+
+# Create a list of the expected output files
+output_files <- c(output_mapped_genes, output_matrix, output_annotation)
+
+# Check that any of the output files exist, only if the overwrite flag is false
+if (!opt$overwrite) {
+  if (any(file.exists(output_files))) {
+    existing_file <- output_files[file.exists(output_files)]
+    print(glue::glue("{existing_file} file already exists."))
```
sjspielman, line 176: I don't think this needs a `print()`?

sjspielman (approved):
Looks good! Just check if you _need_ that print.

optional-goi-analysis/goi-calculations.R
cbethell: I added it here because the only that was printed out otherwise was what's in the `stop()` function -- I double-checked that this is case and indeed it is.

PR #292, opened 2023-02-15
Sync core analysis results files to S3

### LINKED ISSUE #277: Sync core analysis example results to S3 and make link readily available

Now that we have the example results files in our `shared/data` folder, we will want to take that example results folder, compress it, and copy to S3 so that we can produce a single link that can be used for downloads. We will then need to include a section in the README on how to download the example results files per https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/276#issuecomment-1371201301.

Adapt the `sync-results.sh` script introduced in PR #276 to sync the example-results files to the S3 bucket.

cbethell:
**Issue Addressed**
Closes #277 

This PR modifies the `utils/sync-results.sh` script to also sync the core analysis example results files from our shared directory to a S3 bucket named `scpca-downstream-analyses`.

The `aws sync` command was added at the end of the script and the S3 bucket defined at the top of the script. I was able to successfully run this script to populate the S3 bucket that was created on AWS.

The main `README` is also updated in this PR to include links to the example results files in the S3 bucket.

- Does the public S3 bucket seem to be properly created with the right permissions? For ease of reference: https://s3.console.aws.amazon.com/s3/buckets/scpca-downstream-analyses?region=us-east-1&tab=objects
- Do the links to the files work for you?
- Does this script seem to address all of issue #277?

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ There are two expected output files thay will be associated with each provided `
     - Clustering that was performed within the workflow
 
 See the [expected output section](#expected-output) for more information on these output files.
-You can also download a ZIP file with examples of the summary HTML report and processed `SingleCellExperiment` objects stored as RDS files [here](https://scpca-references.s3.amazonaws.com/example-data/scpca-downstream-analyses/example_results.jpg).
+You can also download a ZIP file with an example of the output from running the core workflow, including the summary HTML report and processed `SingleCellExperiment` objects stored as RDS files [here](https://scpca-references.s3.amazonaws.com/example-data/scpca-downstream-analyses/core_example_results.zip).
 
 ## 1. How to install the core downstream analyses workflow
 
@@ -310,7 +310,7 @@ The `<library_id>_core_analysis_report.html` file is the [html file](https://boo
 
 The `<library_id>_processed_sce.rds` file is the [RDS file](https://rstudio-education.github.io/hopr/dataio.html#saving-r-files) that contains the final processed `SingleCellExperiment` object (which contains the filtered, normalized data and clustering results).
 
-You can also download a ZIP file with examples of the summary HTML report and processed `SingleCellExperiment` objects stored as RDS files [here](https://scpca-references.s3.amazonaws.com/example-data/scpca-downstream-analyses/example_results.jpg).
+You can also download a ZIP file with an example of the output from running the core workflow, including the summary HTML report and processed `SingleCellExperiment` objects stored as RDS files [here](https://scpca-references.s3.amazonaws.com/example-data/scpca-downstream-analyses/core_example_results.zip).
 
 ### What to expect in the output `SingleCellExperiment` object
 
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ There are two expected output files thay will be associated with each provided `
     - Clustering that was performed within the workflow
 
 See the [expected output section](#expected-output) for more information on these output files.
-You can also download an example of the summary HTML report [here](https://scpca-downstream-analyses.s3.amazonaws.com/example-results/sample01/library01_core_analysis_report.html) and an example of the processed `SingleCellExperiment` object as a RDS file [here](https://scpca-downstream-analyses.s3.amazonaws.com/example-results/sample01/library01_processed_sce.rds).
+You can also download a ZIP file with examples of the summary HTML report and processed `SingleCellExperiment` objects stored as RDS files [here](https://scpca-references.s3.amazonaws.com/example-data/scpca-downstream-analyses/example_results.jpg).
 
 ## 1. How to install the core downstream analyses workflow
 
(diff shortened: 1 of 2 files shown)


utils/sync-results.sh
```diff
@@ -33,3 +34,5 @@ do
     echo "${loc} already exists and is not a link, delete or move it to create a link."
   fi
 done
+
+aws s3 sync $modules_base $s3_base
```
allyhawkins, line 38:
```suggestion
# zip and sync example results
zip -r core_example_results.zip $repo_base/example_results
aws s3 cp example_results.zip $s3_base/example_results.jpg --acl public-read
```
I believe you want to do something like this instead where we zip up the results directory into a single zip file and then copy that file to S3 so that users can use a single link to download an example of the results directory. You also want to make sure that file is publicly accessible.

utils/sync-results.sh
```diff
@@ -11,6 +11,7 @@ cd ..
 # location for the shared data
 share_base=/shared/data
 modules_base=${share_base}/scpca-downstream-analyses
+s3_base=s3://scpca-downstream-analyses
```
allyhawkins, line 14:
```suggestion
s3_base=s3://scpca-references/example-data/scpca-downstream-analyses
```
Let's actually put these in our public bucket for scpca where we already have stored example results files. I think this makes more sense instead of starting a new bucket.

utils/sync-results.sh
allyhawkins: You probably want to add a step that removes the local copy of the zip file too after copying.

allyhawkins (commented):
My first few comments are on making sure that we have the files present on S3 in a way that we can easily link to them. I think we want to have one link that downloads a zip file with the example results. That way then can see the file structure and open individual files if they would like. I added some suggestions to the script to help do that automatically that we should update and then the text can be updated.

cbethell:
Thanks for the feedback @allyhawkins! I've renamed the repo variable, zipped and copied over the folder as suggested, and stored it within the scpca-references bucket rather than in a separate bucket (I also deleted the new bucket that I had created)! This is ready for another look!

utils/sync-results.sh
```diff
@@ -33,3 +34,8 @@ do
     echo "${loc} already exists and is not a link, delete or move it to create a link."
   fi
 done
+
+# zip and sync example results
+zip -r core_example_results.zip $repo_base/example-results
+aws s3 cp core_example_results.zip $s3_base/example_results.jpg --acl public-read
```
allyhawkins, line 40:
```suggestion
aws s3 cp core_example_results.zip $s3_base/core_example_results.zip --acl public-read
```
This should be `zip` not `jpg`, also let's name it with the full folder name and the core appended to it. I think my thought process here is that we would keep separate zip files with the output from just the core workflow and then each module, otherwise users might get confused seeing all the other output together.

README.md
```diff
@@ -88,6 +88,7 @@ There are two expected output files thay will be associated with each provided `
     - Clustering that was performed within the workflow
 
 See the [expected output section](#expected-output) for more information on these output files.
+You can also download a ZIP file with examples of the summary HTML report and processed `SingleCellExperiment` objects stored as RDS files [here](https://scpca-references.s3.amazonaws.com/example-data/scpca-downstream-analyses/example_results.jpg).
```
allyhawkins, line 91:
```suggestion
You can also download a ZIP file with an example of the output from running the core workflow, including the summary HTML report and processed `SingleCellExperiment` objects stored as RDS files [here](https://scpca-references.s3.amazonaws.com/example-data/scpca-downstream-analyses/core_example_results.zip).
```
Just a minor rewording suggestion, and renaming the file accordingly based on my comment in the actual script.

README.md
```diff
@@ -309,6 +310,8 @@ The `<library_id>_core_analysis_report.html` file is the [html file](https://boo
 
 The `<library_id>_processed_sce.rds` file is the [RDS file](https://rstudio-education.github.io/hopr/dataio.html#saving-r-files) that contains the final processed `SingleCellExperiment` object (which contains the filtered, normalized data and clustering results).
 
+You can also download a ZIP file with examples of the summary HTML report and processed `SingleCellExperiment` objects stored as RDS files [here](https://scpca-references.s3.amazonaws.com/example-data/scpca-downstream-analyses/example_results.jpg).
```
allyhawkins, line 313:
```suggestion
You can also download a ZIP file with an example of the output from running the core workflow, including the summary HTML report and processed `SingleCellExperiment` objects stored as RDS files [here](https://scpca-references.s3.amazonaws.com/example-data/scpca-downstream-analyses/core_example_results.zip).
```

allyhawkins (commented):
These changes look good, just make sure that the file is stored as a zip file on S3. Also we should have a zip file for the example files from each workflow. So the clustering workflow would have a separate zip download and the files should be named with something to distinguish them in s3.

utils/sync-results.sh
cbethell: Ah yes, that makes total sense 👍

cbethell:
_Now_ I think this is just about ready after incorporating those last set of changes!

allyhawkins (approved):
Looks good! And I checked the file in the bucket and it's set to public so look like everything worked as it should 👍

Issue #299, opened 2023-02-22
Thought: Convert from renv to conda-native R package handling


jashapiro:
**Please provide some background on the proposed additions or changes.**

The interaction between `conda` and `renv` seems to be a bit brittle, and while `renv` is great for development, for deployments it may be less than ideal, as building packages can be slow, and occasionally fails.

**What are the changes that you are proposing?**

[Bioconda](https://bioconda.github.io/) now has support for just about all of the packages that we use for this workflow, so we could specify all of the packages in the `scpca-renv.yaml` file and remove the post-deploy script that installs packages via renv.

**Please describe the proposed solution.**

Add to `scpca-renv.yaml` all of the packages, with versions, that we use for this workflow. This can be done by searching the [Bioconda package list](https://bioconda.github.io/conda-package_index.html) for the appropriate `r-*` and `bioconductor-*` packages. 

If we keep the `renv` installation in parallel would probably need to rewrite the `snakemake` rules to be sure that they are using `R --vanilla` to ensure that we are using the conda-based R packages and not the ones that `renv` is expecting. 

**What potential "gotchas" do we know of?**

Keeping the versions in sync between the renv  and conda yaml file may be a challenge... I can't think of an easy solution here other than vigilance.

There may also be packages that are not available in bioconda that I have not yet found, but it seems like most of bioconductor is available there, so I expect this is possible.

allyhawkins:
Closed by #322

PR #300, opened 2023-02-23
Potential Conda/renv 

jashapiro:
**Issue Addressed**

There were problems building the conda environment, which I ultimately tracked down to the fact that the latest version of `purrr` seems to now require `glue`. This only matters in the initial setup, but meant that other things didn't build as expected.  (Note that we were not pinning the `purrr` version here, because it was only used for installation, not in the actual running of code) In debugging I found that building the Apple Silicon versions of some packages (rhdf5) was also failing now, so I went back to using rosetta.

The main change is to load the whole tidyverse, rather than individual packages, when preparing to run the `renv::restore`. This is probably overkill, but it works. To enable this, I also install the `tidyverse` packages via `conda`, which allows me to pin the version, which hopefully will avoid a repeat of this problem in the future. 

Unfortunately, pinning the tidyverse version also seemed to require an updated to R 4.2.2, which seemed like it was probably an accepatable compromise

A potential alternative remains usign conda for all installs as described in #299, but I think that will take longer to implement, and may cause future headaches with keeping the `renv` in sync. 

I'm filing this as a draft in part because I was not sure which branch I should target with these changes. Right now it is targeted at main, but let me know if another target is better! 

Also, others will need to pull these changes to test that things build as expected!

**Checklist**

### DIFF
--- a/envs/scpca-renv.post-deploy.sh
+++ b/envs/scpca-renv.post-deploy.sh
@@ -3,8 +3,7 @@
 # install packages required for renv
 Rscript --vanilla -e \
   "
-   .libPaths(.Library)
-   install.packages(c('jsonlite', 'purrr'), repos = 'https://cloud.r-project.org')
+   library(glue)
    library(jsonlite)
    library(purrr)
    source('.Rprofile')
--- a/envs/scpca-renv.post-deploy.sh
+++ b/envs/scpca-renv.post-deploy.sh
@@ -7,6 +7,4 @@ Rscript --vanilla -e \
    library(jsonlite)
    source('.Rprofile')
    renv::restore()
-  " \
-  &> .snakemake/log/renv_install.log
-
+  " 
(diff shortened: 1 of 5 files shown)


allyhawkins:
Thank you for helping to tackle this and filing this PR! I went to test this on the server and ran into the error that it couldn't find `jsonlite`. So I added `r-jsonlite` and `library(jsonlite)` to the `.yaml` and post deploy script respectively. That then led to the failed installation of `Rhdf5lib`, which it sounds like you had also seen when you were trying to do locally? I'm going to see what I can find and try other options, but thought I would post this here in the meantime, in case there were other things you tried that I can potentially try for building this on the server. 
Screen Shot 2023-02-23 at 9 56 45 AM

jashapiro:
Argh. I thought we were almost there. Let me run more tests on the server.

cbethell:
Interestingly this did work for me locally (fwiw), but no luck on the server as mentioned above.

PR #306, opened 2023-03-03
Use conda only(mostly) to handle R dependencies

jashapiro:
**Issue Addressed**

#299 

We had run into trouble with some of the `renv` commands working in the context of conda environments, so this PR is an attempt to remove (mostly) our use of `renv` from the building of environments.

The main change is to the `scpca-renv.yaml` file, which now includes specification of many more R packages. I created this file by creating an environment where I could manually install R packages using `mamba install r-*` or `mamba install bioconductor-*` until I had installed most of the packages that I could find used in the R scripts for this repository. I then used `conda env export > scpca-renv.yml` to create a record of all of the packages that were installed in that environment. For this initial setup, I did not include version numbers in the installation commands.

This environment file contained both R and Bioconductor packages and all of their system dependencies. Since the system dependencies will vary by platform, I removed everything that was not an `r-*` or `bioconductor-*` package, with the exception of `pandoc`, which I also included. 

The environment file also contains build ids, which can be system-specific, so I stripped all additional build ids (the second `=` and following for each package), leaving only version numbers.

In testing I found a few more packages that I had missed, so I added those to the file manually: I would start a `mamba install <package>` command, but then just look to see what version of the package was going to be installed (to ensure it was compatible with the current packages installed), then updated the `scpca-renv.yaml` file to include that.

This covered CRAN & Bioconductor packages, but we still needed to install `scpcaTools`, so that moved over to the post-install script.

Overall the whole thing runs much faster, so I removed the statement from the README about it taking an hour to set up!

Since we are no longer using `renv`, I also removed the `.Rprofile` file and thew `renv` director, as well as the calls to `setup_renv()`

I did, however, leave the `renv.lock` file. I changed a rule in the workflow to allow us to write out an `renv.lock` file that corresponds to the currently installed (and used) packages. This allows users to continue to use `renv` if they should wish to, though I have not really tested that way of working. A few more thoughts on that below.

First, does it work for you? I tested on my machine and on the server, and things seemed to go well, but I have no confidence that will apply to others. My one tip is that it may be worth deleting `.snakemake/conda` before running `setup-envs.sh` and maybe we should just make that part of the script?

Second, there is still a potential issue with `renv` compatibility. If somebody decides they want to use `renv` without conda, I would expect that to work okay: they could just open R, run `renv::activate()` and `renv::restore()` and things should work since the `renv.lock` file will have the same versions of packages. However, that will create an `.Rprofile` file, and I worry that might cause some trouble if they _afterward_ try to run in the conda environment. I have not yet tested this scenario.

**Still to do:**

Update documentation, particularly about how to add packages, as we probably don't want to do that with a usual `install.packages()` call.

Test the renv/conda combination conundrum.

**Checklist**


allyhawkins:
Thank you so much for doing this! This worked for me on the server 🎉 

My only comment is that we want to get this into the repo without making any other changes for eval purposes so it should be based on `main`. I think it's fine to leave it on `development`, but then we are going to want to cherry pick these changes to merge into `main` so as not to sync all the other changes into `main`. I just wanted to point that out, in case you wanted to re-base now.

cbethell:
I also wanted to note here that I was able to successfully run the `setup-envs.sh` script and the core workflow with the current changes in this PR (after removing the `.snakemake/conda` directory as suggested above). 👍

jashapiro:
closing in favor of #316

PR #317, opened 2023-03-09
Update miQC filtering - remove cells with low gene count

### LINKED ISSUE #308: Update miQC filtering to also remove cells with low gene count

The core workflow has two separate filtering modes, manual or miQC. However in `scpca-nf` when we do filtering we use a combination of miQC and removing cells below a minimum gene cutoff. We should do the same here, so if miQC is used, we should also remove cells with a low number of genes detected. We should also use the same default number for minimum number of genes. 

Additionally, we store the information for filtering slightly differently in `scpca-nf` vs. in the core workflow here. We probably want to make the changes to the metadata here so that they match. Even though we don't use them downstream, I think it would be important for things to be in sync. 

In addition to removing cells that do not pass miQC, cells with a total number of genes below the minmum number of genes detected should be removed during the filter step. 

In the `01-filter-sce.R` script, if the filtering method chose is `miQC` then we should also filter using the `detected` column in the `colData`. This should be done after using the `miQC::filterCells` function. 

Additionally, the default value for `detected_gene_cutoff` should be changed to 200 to match the value in `scpca-nf`. 

The metadata will also need to be updated, specifically the filtering method should be stored as `scpca_filter_method` and either `Minimum_gene_cutoff` or `miQC`. We also want to chnage `detected_gene_cutoff` to be `min_gene_cutoff` and store it for both filtering methods. 

cbethell:
**Issue Addressed**
Closes #308 

This PR adds a step to remove cells with low gene count when miQC filtering is implemented, to sync with what's done in `scpca-nf`.

This PR updates the miQC steps of the core workflow to also include filtering cells with low gene count using the `detected` column in the `colData`, as done in the scpca-nf workflow. This update was made in the `core-analysis/01-filter-sce.R` script.

As also noted on the issue, this PR changes `detected_gene_cutoff` to be `min_gene_cutoff`, changes the default to be 200 and stores it for both filtering methods.

This filtering method is also stored as `scpca_filter_method` now instead of `filtering`, with the two options being `Minimum_gene_cutoff` or `miQC`

Does this PR seem to adequately address #308? Is anything missing?

**Checklist**

### DIFF
--- a/core-analysis/01-filter-sce.R
+++ b/core-analysis/01-filter-sce.R
@@ -83,10 +83,10 @@ option_list <- list(
     metavar = "double"
   ),
   optparse::make_option(
-    c("-p", "--detected_gene_cutoff"),
+    c("-p", "--min_gene_cutoff"),
     type = "integer",
-    default = 500,
-    help = "cell detected genes cutoff -- not needed for miQC filtering",
+    default = 200,
+    help = "cell detected genes cutoff",
     metavar = "integer"
   ),
   optparse::make_option(
@@ -196,6 +196,7 @@ if (opt$filtering_method == "manual") {
                                         mito_percent_cutoff = opt$mito_percent_cutoff,
                                         detected_gene_cutoff = opt$detected_gene_cutoff,
                                         umi_count_cutoff = opt$umi_count_cutoff)
+  metadata(filtered_sce)$scpca_filter_method <- "Minimum_gene_cutoff"
 
 
 } else if (opt$filtering_method == "miQC") {
@@ -217,12 +218,15 @@ if (opt$filtering_method == "manual") {
                           model = model,
                           posterior_cutoff = opt$prob_compromised_cutoff,
                           verbose = FALSE)
-
+      
+      # filter detected genes
+      filtered_sce <- filtered_sce[, colData(filtered_sce)$detected >= opt$min_gene_cutoff]
+      
       # Save model in metadata for plotting later
       metadata(filtered_sce)$miQC_model <- model
 
       # Include note in metadata re: filtering
-      metadata(filtered_sce)$filtering <- "miQC filtered"
+      metadata(filtered_sce)$scpca_filter_method <- "miQC"
       metadata(filtered_sce)$probability_compromised_cutoff <- opt$prob_compromised_cutoff
 
     }, silent = TRUE)
@@ -238,8 +242,9 @@ if (opt$filtering_method == "manual") {
     # manually filter the cells
     filtered_sce <- manual_cell_filtering(sce = sce_qc,
                                           mito_percent_cutoff = opt$mito_percent_cutoff,
-                                          detected_gene_cutoff = opt$detected_gene_cutoff,
+                                          detected_gene_cutoff = opt$min_gene_cutoff,
                                           umi_count_cutoff = opt$umi_count_cutoff)
+    metadata(filtered_sce)$scpca_filter_method <- "Minimum_gene_cutoff"
 
   }
 }
@@ -263,6 +268,7 @@ metadata(filtered_sce)$genes_filtered <- opt$filter_genes
 metadata(filtered_sce)$sample <- opt$sample_id
 metadata(filtered_sce)$library <- opt$library_id
 metadata(filtered_sce)$num_filtered_cells_retained <- dim(filtered_sce)[2]
+metadata(filtered_sce)$min_gene_cutoff <- opt$min_gene_cutoff
 
 # Save output filtered sce
 readr::write_rds(filtered_sce, output_file)
--- a/core-analysis/core-analysis-report-template.Rmd
+++ b/core-analysis/core-analysis-report-template.Rmd
@@ -97,12 +97,12 @@ sample_information <- tibble::tibble(
     format(num_cells_pre_processed, big.mark = ',', scientific = FALSE),
   "Cells after filtering" = 
     format(num_filtered_cells, big.mark = ',', scientific = FALSE),
-  "Filtering type" = 
-    format(processed_meta$filtering, big.mark = ',', scientific = FALSE)
+  "Filtering type" =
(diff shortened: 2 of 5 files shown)


core-analysis/01-filter-sce.R
```diff
@@ -196,6 +196,7 @@ if (opt$filtering_method == "manual") {
                                         mito_percent_cutoff = opt$mito_percent_cutoff,
                                         detected_gene_cutoff = opt$detected_gene_cutoff,
                                         umi_count_cutoff = opt$umi_count_cutoff)
+  metadata(filtered_sce)$scpca_filter_method <- "Minimum_gene_cutoff"
```
allyhawkins, line 199:
```suggestion
  metadata(filtered_sce)$scpca_filter_method <- "manual"
```
So in going back over this, what I think we want to do is keep this here, changing it back to just `manual`, but we should remove the line in the `manual_cell_filtering` function that adds the filtering method to the metadata. 
https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/85df80e018e47f04b0158cd0a4425a6dacf36667/utils/filtering-functions.R#L31

core-analysis/01-filter-sce.R
```diff
@@ -196,6 +196,7 @@ if (opt$filtering_method == "manual") {
                                         mito_percent_cutoff = opt$mito_percent_cutoff,
                                         detected_gene_cutoff = opt$detected_gene_cutoff,
```
allyhawkins, line 197:
```suggestion
                                        min_gene_cutoff = opt$min_gene_cutoff,
```
I also think this should be changed to `min_gene_cutoff` to match the change in argument name.

core-analysis/01-filter-sce.R
```diff
@@ -238,8 +242,9 @@ if (opt$filtering_method == "manual") {
     # manually filter the cells
     filtered_sce <- manual_cell_filtering(sce = sce_qc,
                                           mito_percent_cutoff = opt$mito_percent_cutoff,
-                                          detected_gene_cutoff = opt$detected_gene_cutoff,
+                                          detected_gene_cutoff = opt$min_gene_cutoff,
                                           umi_count_cutoff = opt$umi_count_cutoff)
+    metadata(filtered_sce)$scpca_filter_method <- "Minimum_gene_cutoff"
```
allyhawkins, line 247:
```suggestion
    metadata(filtered_sce)$scpca_filter_method <- "manual"
```

allyhawkins: Just note that this change should be made throughout.

core-analysis/core-analysis-report-template.Rmd
```diff
@@ -97,12 +97,12 @@ sample_information <- tibble::tibble(
...
+    format(processed_meta$scpca_filter_method, big.mark = ',', scientific = FALSE)
 )
 
-if (processed_meta$filtering == "manually filtered"){
-  
+if (processed_meta$scpca_filter_method == "Minimum_gene_cutoff"){
+
   sample_information <- sample_information %>%
     mutate(
       "UMI count cutoff" =
```
allyhawkins, line 108: I can't comment on this line below, but I think we want to change `Detected genes cutoff` to be `Minimum gene cutoff`, and then also include it for the miQC table as well.

utils/filtering-functions.R
```diff
@@ -30,7 +30,6 @@ manual_cell_filtering <- function(sce,
   # Include note in metadata re: filtering
   metadata(filtered_sce)$filtering <- "manually filtered"
```
allyhawkins, line 31: I didn't realize you changed these functions when I commented above, but I would either change this here to `scpca_filter_method` and then remove the two lines you have to add it in in the script, or remove this all together.

allyhawkins (commented):
Thanks for tackling this. I have a few comments on making sure everything is still in sync regarding the naming of filtering methods. Also make sure that we've changed `detected_gene_cutoff` to `min_gene_cutoff` throughout, including the filtering functions and the QC report.

cbethell:
Sweet I've reverted the `Minimum gene cutoff` note to be `manual` and propagated the update where necessary, and successfully tested things with both miQC and manual filtering 👍 Let me know what you think now @allyhawkins!

core-analysis/01-filter-sce.R
```diff
@@ -238,8 +242,9 @@ if (opt$filtering_method == "manual") {
     # manually filter the cells
     filtered_sce <- manual_cell_filtering(sce = sce_qc,
                                           mito_percent_cutoff = opt$mito_percent_cutoff,
-                                          detected_gene_cutoff = opt$detected_gene_cutoff,
+                                          detected_gene_cutoff = opt$min_gene_cutoff,
```
allyhawkins, line 245:
```suggestion
                                          min_gene_cutoff = opt$min_gene_cutoff,
```

core-analysis/core-analysis-report-template.Rmd
```diff
@@ -97,25 +97,25 @@ sample_information <- tibble::tibble(
     format(num_cells_pre_processed, big.mark = ',', scientific = FALSE),
   "Cells after filtering" = 
     format(num_filtered_cells, big.mark = ',', scientific = FALSE),
-  "Filtering type" = 
-    format(processed_meta$filtering, big.mark = ',', scientific = FALSE)
+  "Filtering type" =
+    format(processed_meta$scpca_filter_method, big.mark = ',', scientific = FALSE)
```
allyhawkins, line 101: I would actually add the min gene cutoff here so you only have to add it once and it's included regardless of filtering type.

allyhawkins (commented):
Just missing a few more spots that need to be updated with the correct name.

cbethell:
Made those updates you suggested (as well an update in the argument name of the manual plotting function) -- successfully tested both miQC and manual filtering once again!

allyhawkins (approved):
Looks good 👍

PR #322, opened 2023-03-15
Move to conda-managed R

### LINKED ISSUE #299: Thought: Convert from renv to conda-native R package handling

jashapiro:
This PR applies the changes from #316 to the development branch. 
It should probably be tested again, just to ensure completeness.

Closes #299


cbethell (approved):
Works and looks good to me!

Issue #313, opened 2023-03-06
Add information to documentation on how to run modules using data from ScPCA


allyhawkins:
**Please provide some background on the proposed additions or changes.**

Although we say that users can use ScPCA data that they have downloaded and run it right through the portal, I don't think we've done a good job of providing explicit instructions on how they would set up the workflows using data from the portal. For example, if they have downloaded data from the portal they will receive a `libraries-metadata.csv` that actually has the sample and library id that they will need to set up their metadata for these workflows. Also the files will already be organized in a way they need. So we can be a bit more explicit about how to set up the workflows. 

**What are the changes that you are proposing?**

We should provide a quickstart guide or some sort of separate documentation that is clearly linked on how to setup the metadata file/ data for using any of the workflows in this repo since the setup is the same. 

**Please describe the proposed solution.**

We should include clear step by step instructions that are easily linked from the main readme on how to set up the workflows when using scpca data. Also, we might think of providing a specific example using scpca data that is already available, like an example metadata file showing the scpca ids and where they would go in the metadata needed to run this workflow. 

**What potential "gotchas" do we know of?**

**Additional context or questions**

One idea that I did think of is actually adding in a prep step to the workflow where if the data is from ScPCA they can provide the directory holding the scpca data and we can grab the metadata file from that directory and then create the file needed to run the modules. But I think that might be a future idea. We would probably still need a way for users to specify either the entire project or a list of sample/ library ids.

cbethell:
Per https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/324#pullrequestreview-1355011602, we will also want to include a section before `Running the workflow` that states what the expected input files are more explicitly.

cbethell:
Addressed by #326

PR #324, opened 2023-03-22
Separate `results_dir` from input dir variable in clustering and GOI modules

### LINKED ISSUE #323: Separate input data directory from results directory variable in clustering and GOI modules

Upon testing one of the samples downloaded from the ScPCA portal, an error was encountered due to the clustering module looking within the results directory for the processed RDS file by default (which works fine in the case of running the core workflow first but not otherwise).

We will want to keep the `results_dir` separate from an `input_data_dir`. This update will need to be made the config file and in the Snakefiles for both the clustering and GOI modules.

cbethell:
**Issue Addressed**
Closes #323

This PR creates a separate `input_data_dir` variable for cases where users are using pre-processed data downloaded directly from the ScPCA portal. 

The clustering and GOI Snakefiles were both modified here to use an `input_data_dir` variable from the config file to specify the path to the `_processed.rds` object, and to use the `results_dir` variable to specify the paths to the output files.
The config file was also updated to include the `input_data_dir` with the default path being `example-results` which works in the case where users use the core analysis for generating the processed.rds object.

I also updated the README file in both modules here to reflect the addition of the `input_data_dir` variable.

Does anything else need to be addressed in this PR to complete #323?

**Checklist**

### DIFF
--- a/cluster.snakefile
+++ b/cluster.snakefile
@@ -34,11 +34,11 @@ rule target:
 
 rule calculate_clustering:
     input:
-        "{basedir}/{library_id}_processed.rds"
+        os.path.join(config["input_data_dir"], "{sample_id}/{library_id}_processed.rds")
     output:
-        sce = "{basedir}/{library_id}_clustered_sce.rds",
-        stats_dir = directory("{basedir}/{library_id}_clustering_stats")
-    log: "logs/{basedir}/{library_id}/calculate_clustering.log"
+        sce = os.path.join(config["results_dir"], "{sample_id}/{library_id}_clustered_sce.rds"),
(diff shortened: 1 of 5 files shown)


allyhawkins (approved):
So generally this looks good, but my one overall comment is that I think we're missing a short section in both of the readme's here about what the expected input files are. I think when we are working on #313 that we should include a section before `Running the workflow` that states what the expected input files are more explicitly.

cbethell:
> we should include a section before `Running the workflow` that states what the expected input files are more explicitly.

Thanks @allyhawkins -- I made a note of the above comment on issue #313 and also made a few more changes to the GOI snakefile paths that I missed the first time around. I've now successfully tested a dataset from the portal with the changes made here!

PR #337, opened 2023-04-17
Initiate merge-sce.R 

### LINKED ISSUE #332: Data integration module: Initiate merge SCE script

We will want to initiate a merge SCE script as the first step of developing a data integration module.

Create an `optional-integration-analysis` subdirectory, and within this directory initiate a `merge-sces.R` script that will look similar to the [merge sces script in the scpca-nf repo ](https://github.com/AlexsLemonade/scpca-nf/blob/development/bin/merge_sces.R). 

In the `merge-sces.R` script, we will want to implement the `scpcaTools::merge_sce_list()` function to merge a list of SCE objects.
The input to this script will be the TSV file defining the library ids which can be used to construct the file paths to the processed SCE objects.

cbethell:
**Issue Addressed**
Closes #332

This PR initiates a merge SCE script as the first step of developing a data integration module.

I first created an `optional-integration-analysis` subdirectory, and within this directory initiated a `merge-sce.R` script that looks similar to the [merge sces script in the scpca-nf repo ](https://github.com/AlexsLemonade/scpca-nf/blob/development/bin/merge_sces.R).

The script takes in the project metadata TSV file, and uses the library IDs listed there to find the SCE objects to be merged.

After a series of checks, the SCE objects are merged using the `scpcaTools::merge_sce_list()` function and the output is stored where the user specifies.

- This should probably be tested to ensure that it works for someone who is not me!
- Are there any checks that should be included that I may be missing her?
- Does this seem to adequately address #332?

**Checklist**


optional-integration-analysis/merge-sce.R
```diff
@@ -0,0 +1,128 @@
...
+  library_ids <- input_metadata$library_id
+}
+
+# Check that there is more than one library id associated with an SCE to be merged
+if(length(library_ids) == 1){
+  stop("Only 1 input file provided, no merging or integration will be performed for this group")
+}
+
+# List of SCE filepaths
+sce_files <- library_ids |> purrr::map(~ file.path(opt$input_directory, paste0(.x, "_processed.rds")))
```
allyhawkins, line 80:
```suggestion
sce_files <- file.path(opt$input_directory, input_metadata$sce_filepath)
```
You shouldn't need the `purrr::map` here. 
But also as I'm looking at this, I think we want to require the file path as a column in the metadata and then pull the files directly from the metadata. The reason is because when downloading data from ScPCA, each processed file is going to be in its own sample folder, where as that isn't the case when going from the core workflow. So we either have to search the input directory recursively for the files, or provide the entire path as input. I think right now, the entire input makes sense.

optional-integration-analysis/merge-sce.R
```diff
@@ -0,0 +1,128 @@
...
+
+# Check that input files exist
+missing_sce_files <- c()
+for (sce in sce_files) {
+  if(!file.exists(sce)){
+    missing_sce_files <- missing_sce_files + sce
+  }
+}
```
allyhawkins, line 88:
```suggestion
missing_sce_files <- sce_files[!which(file.exists(sce_files))]
```
You can simplify this a lot

optional-integration-analysis/merge-sce.R
```diff
@@ -0,0 +1,128 @@
...
+  stop("output file name must end in .rds")
+}
+
+# Check that input metadata file is provided
+if(is.null(opt$input_metadata_tsv)){
+  stop("TSV file with the list of library IDs associated with the individual SCE objects to merge is missing.")
+} else {
+  # List of library ids
+  input_metadata <- readr::read_tsv(opt$input_metadata_tsv)
+  library_ids <- input_metadata$library_id
```
allyhawkins, line 71:
```suggestion
```
I actually don't think we need the library IDs, if we are going to explicitly grab the file paths. 
We should eventually include the library ID and a grouping column in the metadata so that we can specify groups of libraries to integrate together, but for now we are going to integrate all files and add the grouping in a separate PR.

optional-integration-analysis/merge-sce.R
cbethell: Good point 👍

cbethell:
Thanks for the review @allyhawkins! I implemented your suggestions, with the biggest modification being adding an expected to column to the input metadata file named `processed_sce_filepath` for this use case -- with the reason being to keep this file path separate from the filtered rds filepath which would be used as input to the core workflow.

Let me know what you think!

optional-integration-analysis/merge-sce.R
```diff
@@ -0,0 +1,108 @@
...
+# Check that input metadata file is provided
+if(is.null(opt$input_metadata_tsv)){
+  stop("TSV file with the list of library IDs associated with the individual SCE objects to merge is missing.")
+} else {
+  # List of library ids
+  input_metadata <- readr::read_tsv(opt$input_metadata_tsv)
+}
+
+# List of SCE filepaths
+sce_files <- input_metadata$processed_sce_filepath
```
allyhawkins, line 69:
```suggestion
sce_files <- input_metadata$processed_sce_filepath |> 
purrr::set_names(input_metadata$library_id)
```
I think you want to set the names of the `sce_files` here to be the `library_id` column. Otherwise the `merge_sce_list` function will print out a warning and not properly identify which library belongs to which SCE object. Also if you name the files here, when you read in the RDS files using `purrr` they should retain the names.

allyhawkins (commented):
I have one remaining comment, but other than that it looks good. Also did you mean to add in the additional metadata column here in this PR? Because it's missing right now.

cbethell:
> I have one remaining comment, but other than that it looks good. Also did you mean to add in the additional metadata column here in this PR? Because it's missing right now.

I initially did not add that additional metadata column because I thought it might confuse those looking at the example metadata file for implementation with other modules -- I wondered if we should create a separate integration example metadata file but I do think the current way should be fine with proper documentation at this point so I added that column in the last commit and I can revert the change if you disagree @allyhawkins!

allyhawkins:
> I initially did not add that additional metadata column because I thought it might confuse those looking at the example metadata file for implementation with other modules -- I wondered if we should create a separate integration example metadata file but I do think the current way should be fine with proper documentation at this point so I added that column in the last commit and I can revert the change if you disagree @allyhawkins!

I think this is a good point! Can we make a separate example integration metadata file? We are going to also eventually add an `integration_group` column too so users can specify which libraries to integrate together so it's probably best to keep it separate in the examples.

cbethell:
> I think this is a good point! Can we make a separate example integration metadata file? We are going to also eventually add an `integration_group` column too so users can specify which libraries to integrate together so it's probably best to keep it separate in the examples.

Okay sweet, I've added that separate metadata file for integration to the `example-data/project_metadata` directory and reverted the changes to the existing metadata file -- let me know what you think of things now!

optional-integration-analysis/merge-sce.R
```diff
@@ -0,0 +1,109 @@
...
+option_list <- list(
+  make_option(
+    opt_str = c("--input_metadata_tsv"),
+    type = "character",
+    help = "Path to the TSV file containing the list of library IDs corresponding to the libraries being integrated, must have `library_id` and `processed_sce_filepath` columns."
+  ),
+  make_option(
+    opt_str = c("-o", "--output_sce_file"),
+    type = "character",
+    help = "Path to output RDS file, must end in .rds"
```
allyhawkins, line 20:
```suggestion
    help = "Path to output RDS file containing merged object, must end in .rds"
```
Just to clarify what exactly is being output here.

allyhawkins (approved):
Looks good, just one minor documentation comment.

PR #342, opened 2023-05-22
Allow for merging of SCEs by integration group

### LINKED ISSUE #340: Data Integration: Allow for merging of SCEs by integration group

Right now we have initiated a script for merging SCE objects that takes as input a metadata file where each processed library file is a row in the metadata file. The script as it works right now will merge all libraries listed in the metadata into one merged SCE object. However, it's likely that users want to be able to integrate specific groups and don't necessarily want one single integrated object.  

We should allow users to specify which group each library belongs to (e.g. libraries from the same disease are merged together rather than all libraries being merged together into one). To do this we will need to add an additional column to the integration library metadata file where users can indicate which libraries should be grouped together. 

This issue will involve two new additions: 
1. Adding a new column, `integration_group` to the integration metadata to indicate which group each library should be added to. 
2. Minimal modifications to workflow so that we are only merging libraries with the same `integration_group` value. Right now we read in the list of sce files to include in a merged object. I believe we should be able to continue to do that and add a `integration_group` argument. Then I would add the `integration_group` to the metadata of the merged SCE object. 

This approach is slightly different than we have in the integration workflow in `sc-data-integration`, where we read in the metadata file and split the data frame by the `integration_group` before reading in the sce objects and merging. 
I think we should be able to have minimal modifications on the merging script, but grab the list of SCE files for each `integration_group` listed in the metadata file within snakemake. 

cbethell:
**Issue Addressed**
Closes #340

The purpose of this PR is to allow users to specify which integration group each SCE belongs to (e.g. libraries from the same disease are merged together rather than all libraries being merged together into one).

The following modifications were made to the `merge-sce.R` script in the data integration module:
- an argument to take the desired `integration_group`
- a step to filter in the input metadata on that integration group argument

These modifications allow users to indicate which libraries (SCEs) should be grouped together.

The example metadata file was also updated to included an `integration_group` column and filled in with generic values.

Does this PR see to adequately address #340?

**Checklist**

### DIFF
--- a/optional-integration-analysis/merge-sce.R
+++ b/optional-integration-analysis/merge-sce.R
@@ -15,11 +15,16 @@ option_list <- list(
     help = "Path to the TSV file containing the list of library IDs corresponding to the libraries being integrated, must have `library_id` and `processed_sce_filepath` columns."
   ),
   make_option(
+    opt_str = c("--integration_group"),
+    type = "character",
+    help = "The integration group associated with the SCEs to be merged"
+  ),
+  make_option(
     opt_str = c("-o", "--output_sce_file"),
     type = "character",
     help = "Path to output RDS file containing merged object, must end in .rds"
   ),
-  optparse::make_option(
+  make_option(
     c("--project_root"),
     type = "character",
     default = NULL,
@@ -70,6 +75,15 @@ if(is.null(opt$input_metadata_tsv)){
 } else {
   # List of library ids
   input_metadata <- readr::read_tsv(opt$input_metadata_tsv)
+}
+
+# Check that integration group is provided
+if(is.null(opt$integration_group)){
+  stop("The name of the corresponding integration group is missing.")
+} else {
+  # Filter the metadata to include only information relevant to the specified integration group
+  input_metadata <- input_metadata %>%
+    dplyr::filter(integration_group == opt$integration_group)
 }
 
 # List of SCE filepaths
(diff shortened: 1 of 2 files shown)


optional-integration-analysis/merge-sce.R
```diff
@@ -72,6 +77,15 @@ if(is.null(opt$input_metadata_tsv)){
   input_metadata <- readr::read_tsv(opt$input_metadata_tsv)
 }
 
+# Check that integration group is provided
+if(is.null(opt$integration_group)){
+  stop("The name of the corresponding integration group is missing.")
```
allyhawkins, line 82:
```suggestion
  stop("The `integration_group` is missing and required for identifying which objects should be merged together.")
```
Just clarifying the error message a little bit.

optional-integration-analysis/merge-sce.R
```diff
@@ -72,6 +77,15 @@ if(is.null(opt$input_metadata_tsv)){
...
+} else {
+  # Filter the metadata to include only information relevant to the specified integration group
+  input_metadata <- input_metadata %>%
+    dplyr::filter(integration_group == opt$integration_group)
```
allyhawkins, line 86: Before we get here we should add a check that `opt$integration_group` is in `input_metadata$integration_group` first.

allyhawkins (commented):
Just a few minor comments that should be addressed.

cbethell:
Thanks for the review @allyhawkins! I added in that extra check and removed the additional ids from the example metadata as suggested -- let me know what you think now.

PR #343, opened 2023-05-24
Initiate integration module snakefile

### LINKED ISSUE #341: Data Integration: Initiate workflow with merging rule

Since we now have the merging script underway, we should get started on setting up the snakemake workflow to make sure our setup will work as we expect. 

Initiate the snakemake workflow and add in the rule to create the merged objects. 

Initiating the workflow should entail reading in the integration metadata and specifying the expected output of the merged SCE files, 1 for each `integration_group`. 
We also will need to have a rule that does the merging and takes as input the list of SCE objects for a given group and the name of the output file. 

This will be dependent on addition of the `integration_group` to the integration metadata file in #340.

cbethell:
**Issue Addressed**
Closes #341

This PR initiates the data integration snakemake workflow by adding `integration.snakefile`.

In this PR,

- The `integration.snakefile` is added with the `merge-sce` rule that runs the first part of the workflow (the `merge-sce.R` script) 
- A `integration_config.yaml` file is added with the variables that need to be supplied to the snakefile
- The `envs/scpca-renv.yaml` file is updated to include the required `batchelor` bioconductor package.

- Does the PR adequately address #341?
- More importantly, can you run the code in this PR? 
There is a new bioconductor package required here, `batchelor` -- after updating `scpca-renv.yaml` and `renv.lock` with this package, I encounter the following error when trying to run `integration.snakefile` and/or `setup_envs.sh`:

```
Encountered problems while solving:
  - nothing provides requested bioconductor-batchelor 1.14.1**
```

Am I maybe missing a step here for adding a new package? cc: @allyhawkins @jashapiro 

**Checklist**

### DIFF
--- a/config/integration_config.yaml
+++ b/config/integration_config.yaml
@@ -1,5 +1,6 @@
 # All parameters included in this file can be altered at the command line using the `--config` flag or by editing this file directly.
 
 ### Optional integration module specific parameters
-project_metadata: "example-data/project-metadata/example-integration-library-metadata.tsv"
-results_dir: "example-results"
+integration_project_metadata: "example-data/project-metadata/example-integration-library-metadata.tsv"
+n_hvg: 2000 # number of high variance genes to use for dimension reduction
+threads: 1 # number of multiprocessing threads to use
--- a/config/integration_config.yaml
+++ b/config/integration_config.yaml
@@ -1,0 +1,5 @@
+# All parameters included in this file can be altered at the command line using the `--config` flag or by editing this file directly.
+
+### Optional integration module specific parameters
+project_metadata: "example-data/project-metadata/example-integration-library-metadata.tsv"
+results_dir: "example-results"
--- a/config/integration_config.yaml
+++ b/config/integration_config.yaml
@@ -2,5 +2,4 @@
 
 ### Optional integration module specific parameters
 integration_project_metadata: "example-data/project-metadata/example-integration-library-metadata.tsv"
-n_hvg: 2000 # number of high variance genes to use for dimension reduction
 threads: 1 # number of multiprocessing threads to use
--- a/envs/scpca-renv.yaml
+++ b/envs/scpca-renv.yaml
@@ -5,7 +5,7 @@ channels:
   - defaults
 dependencies:
   - bioconductor-annotationdbi=1.60.0
-  - bioconductor-batchelor=1.16.0
+  - bioconductor-batchelor=1.14.1
   - bioconductor-beachmat=2.14.0
   - bioconductor-biobase=2.58.0
   - bioconductor-biocgenerics=0.44.0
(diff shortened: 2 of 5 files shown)


envs/scpca-renv.yaml
```diff
@@ -5,6 +5,7 @@ channels:
   - defaults
 dependencies:
   - bioconductor-annotationdbi=1.60.0
+  - bioconductor-batchelor=1.14.1
```
jashapiro, line 8: It looks like the latest version available on bioconda is 1.14.0
```suggestion
  - bioconductor-batchelor=1.14.0
```
https://bioconda.github.io/recipes/bioconductor-batchelor/README.html?highlight=batchelor#package-package%20&#x27;bioconductor-batchelor&#x27;

config/integration_config.yaml
```diff
@@ -0,0 +1,5 @@
+# All parameters included in this file can be altered at the command line using the `--config` flag or by editing this file directly.
+
+### Optional integration module specific parameters
+project_metadata: "example-data/project-metadata/example-integration-library-metadata.tsv"
```
allyhawkins, line 4: Hmm I think we might want to actually change this to be `integration_project_metadata`? That way it's distinct from the metadata file that is for the other workflows and we can still use our main config file.

integration.snakefile
```diff
@@ -0,0 +1,37 @@
+import pandas as pd
+
+configfile: "config/integration_config.yaml"
```
allyhawkins, line 3: If you take my earlier suggestion, then you can use both configs here and grab the results directory from that. 
```suggestion
configfile: "config/config.yaml"
configfile: "config/integration_config.yaml"
```

config/integration_config.yaml
```diff
@@ -0,0 +1,5 @@
...
+results_dir: "example-results"
```
allyhawkins, line 5:
```suggestion
```
To be consistent with the other workflows, we should use the results directory we define in the main config file and keep the module config file specific to any parameters unique to integration.

integration.snakefile
```diff
@@ -0,0 +1,37 @@
...
+
+# getting the samples information
+if os.path.exists(config['project_metadata']):
+  samples_information = pd.read_csv(config['project_metadata'], sep='\t', index_col=False)
+
+  # get a list of the file paths and integration groups
+  FILEPATHS = list(samples_information['processed_sce_filepath'])
```
allyhawkins, line 10:
```suggestion
```
I don't think you need this actually, because we don't reference it anywhere in the snakemake workflow. The only time the file paths are used is in the Rscript.

allyhawkins (commented):
This mostly looks good, I just had a few suggestions for cleaning up the config file and keeping things consistent with other modules.

envs/scpca-renv.yaml
cbethell: Ahh! Yes that did the trick, thank you 🎉

cbethell:
Thanks @allyhawkins for the refactoring suggestions, I've implemented them and also added a couple parameters to the `integration_config.yaml` file that I missed the first time around. I have successfully run things with these new updates so this is ready for another look!

allyhawkins (approved):
Looks good 👍 just one minor comment

PR #348, opened 2023-06-07
Incorporate integration order

### LINKED ISSUE #347: Add integration order to integration script

When providing a merged object to the integration script in #346, no integration order is specified. This will not matter for `harmony`, but it will matter for `fastMNN` as the default is to merge in the input order. We did some testing in `sc-data-integration` regarding integration order, and found that the `auto.merge` option in `fastMNN` worked well and we would like to be able to implement that when we implement integration. 

We will want to be able to use the `auto.merge` option for `fastMNN` integration, but also allow for a user to decide they would like to use the input order for integration. This means adding a new argument to account for integration order specifically for `fastMNN` to the script being added in #346. 
We also need to add the `merge.order` option which will allow for users to specify the merge order if different from the input integration order. 

We need to add an option for using `auto.merge` to the script like the one found in `sc-data-integration`: 
https://github.com/AlexsLemonade/sc-data-integration/blob/3994b1934e255a129d5efcb56fa5fa2130d075dc/scripts/03a-integrate-sce.R#L142-L147

This argument should be added to the script in #346 and then implemented with `auto.merge = opt$fastmnn_auto_merge`. 
We will also need to account for this argument in the snakefile and set the default to use `auto.merge`, with the option of turning it off. 

Alternatively, if users want to use a specified order for their libraries, we should create an additional argument for `merge.order`. This should only be used if `auto.merge` is set to false and will have a vector of library id's in the order of which they should be merged. 

My one concern with allowing for the merge order as input is if we have multiple groups that are being integrated, the merge order will need to be specified for each group. This will only be the case if that option is being used, but we should think about how we would want to implement that (maybe a separate column for the order in the metadata?).

cbethell:
**Issue Addressed**
Closes #347

This PR allows us to use the `auto.merge` option for fastMNN integration, and also allows for users to provide a character vector of library ids to use as the order for integration. 

The following modifications were made to the `perform-integration.R` script:
- The `--fastmnn_auto_merge` argument added (default is currently false, but will be set to true using the `config.yaml` file to set the argument in the snakemake workflow in an upcoming PR)
- The `fastmnn_merge_order` argument added and logic to use this list when provided (only if `--fastmnn_auto_merge` is set to false, otherwise this argument will be NULL) 

The formatting of the `fastmnn_merge_order` at line 108 was implemented because without this, I receive the following error:

```
Error in .create_tree_predefined(batches, restrict, merge.order) : 
  invalid leaf nodes specified in 'merge.order'
```

Otherwise, users would need to use `c("library02", "library01")` when implementing this argument. 
- Does this PR correctly address what's noted on issue #347?

**Checklist**

### DIFF
--- a/optional-integration-analysis/perform-integration.R
+++ b/optional-integration-analysis/perform-integration.R
@@ -28,13 +28,14 @@ option_list <- list(
     action = "store_true",
     default = FALSE,
     help = "Indicates whether or not to use the auto.merge option for `fastMNN` integration;
-    to perform auto.merge, use `--fastmnn_auto_merge`"
+    to perform auto.merge, use `--fastmnn_auto_merge` and this will override any input order"
   ),
   make_option(
     opt_str = c("--fastmnn_merge_order"),
     type = "character",
     default = NULL,
-    help = "Optional vector of library ids in the order of which they should be merged"
+    help = "Optional vector of library ids in the order of which they should be merged;
+    will only be used if --fastmnn_auto_merge is FALSE"
   ),
   make_option(
     opt_str = c("-o", "--output_sce_file"),
@@ -103,11 +104,16 @@ if(!("library_id" %in% colnames(colData(merged_sce)))){
 # Perform integration with specified method
 if ("fastMNN" %in% integration_methods) {
   # Format `fastmnn_merge_order` if provided; can only be used when auto.merge is FALSE
-  if(is.null(opt$fastmnn_auto_merge)) {
+  if(!opt$fastmnn_auto_merge) {
     if(!is.null(opt$fastmnn_merge_order)){
       fastmnn_merge_order <- stringr::str_split(opt$fastmnn_merge_order, ",") %>%
         unlist() %>%
         stringr::str_trim()
+      if(!all(colData(merged_sce)$library_id %in% fastmnn_merge_order)){
+        stop("The provided library ids do not match those in the SCE object. 
+             Please re-run and provide --fastmnn_merge_order with all library ids 
+             found in the `colData` of the SCE object.")
+      }
     }
   } else {
     fastmnn_merge_order <- NULL
--- a/optional-integration-analysis/perform-integration.R
+++ b/optional-integration-analysis/perform-integration.R
@@ -24,6 +24,20 @@ option_list <- list(
     default is 'fastMNN, harmony'"
   ),
   make_option(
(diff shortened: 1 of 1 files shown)


optional-integration-analysis/perform-integration.R
```diff
@@ -23,6 +23,19 @@ option_list <- list(
...
+    action = "store_true",
+    default = FALSE,
+    help = "Indicates whether or not to use the auto.merge option for `fastMNN` integration;
+    to perform auto.merge, use `--fastmnn_auto_merge`"
+  ),
+  make_option(
+    opt_str = c("--fastmnn_merge_order"),
+    type = "character",
+    default = NULL,
+    help = "Optional vector of library ids in the order of which they should be merged"
```
allyhawkins, line 37: I think we should add a note here in the help message about how this will only be used if `fastmnn_auto_merge` is false. And then for the `fastmnn_auto_merge` argument we should note that using this will override any input order.

optional-integration-analysis/perform-integration.R
```diff
@@ -89,9 +102,21 @@ if(!("library_id" %in% colnames(colData(merged_sce)))){
 
 # Perform integration with specified method
 if ("fastMNN" %in% integration_methods) {
+  # Format `fastmnn_merge_order` if provided; can only be used when auto.merge is FALSE
+  if(is.null(opt$fastmnn_auto_merge)) {
```
allyhawkins, line 106:
```suggestion
  if(!opt$fastmnn_auto_merge) {
```
Since the default is set to false, I think you should be able to check for false rather than it being null.

optional-integration-analysis/perform-integration.R
```diff
@@ -89,9 +102,21 @@ if(!("library_id" %in% colnames(colData(merged_sce)))){
...
+    if(!is.null(opt$fastmnn_merge_order)){
+      fastmnn_merge_order <- stringr::str_split(opt$fastmnn_merge_order, ",") %>%
+        unlist() %>%
+        stringr::str_trim()
```
allyhawkins, line 110: Let's add a check after this to make sure that the input ID's provided here are found in the merged SCE object. They also should probably encompass all of the libraries present in the merged object otherwise I would assume `fastMNN` would fail.

allyhawkins (commented):
This mostly looks good and is the right idea. I think we are just missing a check.

PR #349, opened 2023-06-08
Add integration rule to `integration.snakefile`

### LINKED ISSUE #345: Data Integration: Add integration rule to Snakefile

Once we have a developed the integration R script (#344), we will want to add a rule to the integration Snakefile to perform integration on each merged object (for each integration group in other words).

Add a `perform_integration` rule to the `integration.snakefile` that runs the script developed in #344.

The second rule added to the `integration.snakefile` should be one named something like `perform_integration` which does the following:
- takes the output of the first `merge_sces` rule as input
- implements the R script responsible for performing integrations (for fastMNN and/or Harmony)
- outputs an integrated RDS object

cbethell:
**Issue Addressed**
Closes #345 

This PR adds a `perform_integration` rule to the data integration snakemake workflow.

This PR makes the following modifications:
- addition of default parameters to the `integration_config.yaml` file need to run the `perform-integration.R` script
- addition of the `perform_integration` rule to the `integration.snakefile` -- which runs the R script responsible for performing integration
- updates the `renv.lock` file and the `envs/scpca-renv.yaml` file with package dependencies

This is still a WIP because I have not yet successfully run the workflow with the added changes -- I am getting the following error after confirming the [right package version in bioconda ](https://anaconda.org/bioconda/r-harmony):

```
Encountered problems while solving:
  - package r-harmony-0.1-r40ha1b3eb9_0 requires r-base >=4.0,<4.1.0a0, but none of the providers can be installed
```

I have also tried changing the r-base version we use but got the same error -- let me know if you have any suggestions here @allyhawkins, otherwise I will tag Josh at the beginning of next week to see if he has any suggestions here.

- Does this PR fully address #345?
- Can you run the snakemake workflow?

**Checklist**

### DIFF
--- a/optional-integration-analysis/perform-integration.R
+++ b/optional-integration-analysis/perform-integration.R
@@ -3,6 +3,10 @@
 
 ## Set up ----------------------------------------------------------------------
 
+if (!("harmony" %in% installed.packages())) {
+  install.packages("harmony", repos = "http://cran.us.r-project.org")
+}
+
 # Load libraries
 library(optparse)
 library(dplyr)
@@ -19,9 +23,9 @@ option_list <- list(
   make_option(
     opt_str = c("--integration_method"),
     type = "character",
-    default = "fastMNN, harmony",
+    default = "fastMNN,harmony",
     help = "The integration method(s) to use when performing integration; 
-    default is 'fastMNN, harmony'"
+    default is 'fastMNN,harmony'"
   ),
   make_option(
     opt_str = c("--fastmnn_auto_merge"),
--- a/optional-integration-analysis/perform-integration.R
+++ b/optional-integration-analysis/perform-integration.R
@@ -4,7 +4,7 @@
 ## Set up ----------------------------------------------------------------------
 
 if (!("harmony" %in% installed.packages())) {
-  install.packages("harmony", repos = "http://cran.us.r-project.org")
+  remotes::install_version("harmony", version = "0.1.1")
 }
 
 # Load libraries
--- a/config/integration_config.yaml
+++ b/config/integration_config.yaml
@@ -3,3 +3,4 @@
 ### Optional integration module specific parameters
 integration_project_metadata: "example-data/project-metadata/example-integration-library-metadata.tsv"
 threads: 1 # number of multiprocessing threads to use
+integration_method: "fastMNN, harmony" # method(s) to be used for integration
--- a/config/integration_config.yaml
+++ b/config/integration_config.yaml
@@ -3,4 +3,4 @@
 ### Optional integration module specific parameters
 integration_project_metadata: "example-data/project-metadata/example-integration-library-metadata.tsv"
 threads: 1 # number of multiprocessing threads to use
-integration_method: "fastMNN, harmony" # method(s) to be used for integration
+integration_method: "fastMNN,harmony" # method(s) to be used for integration
--- a/envs/scpca-renv.yaml
+++ b/envs/scpca-renv.yaml
@@ -145,7 +145,7 @@ dependencies:
   - r-gtable=0.3.1
   - r-gtools=3.9.4
   - r-hardhat=1.2.0
-  - r-harmony=0.1.1
+  - r-harmony=0.1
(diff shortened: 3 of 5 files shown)


cbethell:
As noted above, this is still a WIP because I have not yet successfully run the workflow with the added changes -- I am getting the following error after confirming the [right package version in bioconda ](https://anaconda.org/bioconda/r-harmony):

```
Encountered problems while solving:
  - package r-harmony-0.1-r40ha1b3eb9_0 requires r-base >=4.0,<4.1.0a0, but none of the providers can be installed
```

I have also tried changing the r-base version we use but got the same error -- let me know if you have any suggestions here @allyhawkins, otherwise I will tag Josh at the beginning of next week to see if he has any suggestions here.

allyhawkins:
> As noted above, this is still a WIP because I have not yet successfully run the workflow with the added changes -- I am getting the following error after confirming the [right package version in bioconda ](https://anaconda.org/bioconda/r-harmony):

Have you tried running `bash setup_envs.sh` to make sure the environment has been updated before running the Snakefile? I don't believe we have a rule that syncs the environment like we do in the main workflow and maybe we should have that here? 
https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/179d65ad29bc53ad5cf9f6af042bc8f5ef9c487f/Snakefile#L31-L39

cbethell:
> > As noted above, this is still a WIP because I have not yet successfully run the workflow with the added changes -- I am getting the following error after confirming the [right package version in bioconda ](https://anaconda.org/bioconda/r-harmony):
> 
> Have you tried running `bash setup_envs.sh` to make sure the environment has been updated before running the Snakefile? I don't believe we have a rule that syncs the environment like we do in the main workflow and maybe we should have that here?
> 
> https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/179d65ad29bc53ad5cf9f6af042bc8f5ef9c487f/Snakefile#L31-L39

I got the same error upon running `bash setup_envs.sh` -- perhaps @jashapiro may have some suggestions at this point

jashapiro:
It looks like for some reason there is no harmony version on bioconda for R 4.2 & macOS. No idea why and I can't find any docs about it.

One solution would be to create a separate conda environment for harmony & only the packages that you need for the integration script. Looking at the scripts though, you might actually have to separate the integration methods into separate scripts, so we would have one that _just_ did the harmony integration. 

Another option is to add installing harmony (if not present) as a part of the integration script. I'm not totally sure how that would integrate with the snakemake-installed packages, but it is probably worth a shot.

optional-integration-analysis/perform-integration.R
```diff
@@ -3,6 +3,10 @@
 
 ## Set up ----------------------------------------------------------------------
 
+if (!("harmony" %in% installed.packages())) {
+  install.packages("harmony", repos = "http://cran.us.r-project.org")
```
jashapiro, line 7: We should probably use a frozen version here, either with setting the `repos` to `https://packagemanager.posit.co/cran/2023-06-13` (or another date) or (maybe better) with `remotes::install_version("harmony", version = "0.1.1")`

This assumes that we have `remotes` installed, but that should work to install via condo if it isn't there already.

cbethell: Incorporated `remotes::install_version("harmony", version = "0.1.1")` in https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/349/commits/7d04bc9b364bfcb95044304d73313d6027d9d2f7 👍

allyhawkins (approved):
So the code here looks fine and works now that I remembered to use the `--use-conda` flag. 

Some thoughts that I think we want to change after doing this review. We should figure out how to make `--use-conda` the default which I think we will need to do by creating a shell script that runs the snakefile and sets that as the default. 
The more I think about making a shell script to run this, the more I like it because then we can actually include the core workflow and then the following modules as optional modules to run after that and hopefully make things actually easier for users to use. 

I also ran into errors with trying to run anything without the `--use-conda` option, which is probably something we might want to address? Unless we want to completely remove the ability to run this without conda.

PR #365, opened 2023-07-05
Make `use-conda` a default

### LINKED ISSUE #360: Make `--use-conda` a default

We should try to make the `--use-conda` flag a default somewhere, since it is required to run any of the workflows in repo and it would be nice to have one less thing to have folks remember to include upon implementation.

We can try one of two things here: 

1. Try to set `use-conda` as default in the main config file
2. Make a shell script that runs the snakefile and has --use-conda as default

This modification will also require us to update the documentation on how to run the workflows

cbethell:
**Issue Addressed**
Closes #360

This PR makes the `use-conda` command line option a default in the main config file.

This PR adds `use-conda:TRUE` to the `config/config.yaml` file which is sourced in each of the modules in addition to the core analysis.

Documentation has also been updated here to reflect that there is no longer a need to use the `--use-conda` flag when running the snakemake workflow(s).

The independent installation instruction documentation was also removed in this PR as they have not been updated in a while and using conda is now the default so we want to encourage users to go the "conda" route.

- Are you able to successfully run any of the snakefiles on your machine with this new modification?
- Any other documentation that should be updated that I may have missed?

**Checklist**


config/config.yaml
```diff
@@ -14,6 +14,9 @@ project_metadata: "example-data/project-metadata/example-library-metadata.tsv"
 # Changes to these are not required and can therefore be left at the default values.
 # Although one may want to alter these parameters for exploratory purposes.
 
+# use conda environment
```
allyhawkins, line 17: I think we probably want to add a comment here warning users not to change this.

optional-clustering-analysis/README.md
```diff
@@ -86,17 +86,15 @@ Learn more about the [processing parameters](../additional-docs/processing-param
```
allyhawkins, line 86: I can't comment on this but the note about dependencies on line 41 needs to be updated.

optional-goi-analysis/README.md
```diff
@@ -90,17 +90,15 @@ Learn more about the [gene mapping parameters](../additional-docs/additional-par
 The execution file with the genes of interest Snakemake workflow is named `goi.snakefile` and can be found in the root directory.
```
allyhawkins, line 90: Same thing here about updating the comment on line 34.

allyhawkins (commented):
This mostly looks good. I just pointed out a few places that we were missing updates.

optional-clustering-analysis/README.md
```diff
@@ -38,10 +38,9 @@ Cluster validity and stability results are also calculated.
...
 The plots are displayed in a html report for ease of reference.
 
-**Note** that the same [software requirements for the core workflow](../README.md#3-additional-dependencies) are also required for this clustering workflow.
-R 4.2 is required for running our pipeline, along with Bioconductor 3.15.
-Package dependencies for the analysis workflows in this repository are managed using [`renv`](https://rstudio.github.io/renv/index.html), which must be installed locally prior to running the workflow.
-If you are using conda, dependencies can be installed as [part of the initial setup](../README.md#snakemakeconda-installation).
+**Note** that R 4.1 is required for running our pipeline, along with Bioconductor 3.14.
+Package dependencies for the analysis workflows in this repository are managed using [`renv`](https://rstudio.github.io/renv/index.html), and `renv` must be installed locally prior to running the workflow.
+These dependencies should be installed as [part of the installation mentioned in step 1 of this repository's main README file](../README.md#snakemakeconda-installation).
```
allyhawkins, line 43: I think this is confusing and we should note that we only want people to use conda: 

```suggestion
**Note** that the same [software requirements for the core workflow](https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/cbethell/make-conda-default/README.md#3-additional-dependencies) are also required for this clustering workflow. 
R 4.2 is required for running our pipeline, along with Bioconductor 3.15. If you are using conda, dependencies can be installed as [part of the initial setup](https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/cbethell/make-conda-default/README.md#snakemakeconda-installation). 
Package dependencies for the analysis workflows in this repository are managed using [renv](https://rstudio.github.io/renv/index.html), which can be installed independently if desired. 
Still, we recommend using Snakemake's conda integration to set up the R environment and all dependencies that the workflow will use.

```
I took this from the integration docs.

optional-goi-analysis/README.md
```diff
@@ -31,10 +31,9 @@ The normalized and transformed expression for each of the provided genes of inte
 UMAP and PCA plots are also provided, where each dot represents a cell and the color indicates the individual gene of interest’s expression.
 
 
-**Note** that the same [software requirements for the core workflow](../README.md#3-additional-dependencies) are also required for this clustering workflow.
-R 4.2 is required for running our pipeline, along with Bioconductor 3.15.
-Package dependencies for the analysis workflows in this repository are managed using [`renv`](https://rstudio.github.io/renv/index.html), which must be installed locally prior to running the workflow.
-If you are using conda, dependencies can be installed as [part of the initial setup](../README.md#snakemakeconda-installation).
+**Note** that R 4.1 is required for running our pipeline, along with Bioconductor 3.14.
+Package dependencies for the analysis workflows in this repository are managed using [`renv`](https://rstudio.github.io/renv/index.html), and `renv` must be installed locally prior to running the workflow.
+These dependencies should be installed as [part of the installation mentioned in step 1 of this repository's main README file](../README.md#snakemakeconda-installation).
```
allyhawkins, line 36: Same suggestion as above. 
```suggestion
**Note** that the same [software requirements for the core workflow](https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/cbethell/make-conda-default/README.md#3-additional-dependencies) are also required for this clustering workflow. 
R 4.2 is required for running our pipeline, along with Bioconductor 3.15. If you are using conda, dependencies can be installed as [part of the initial setup](https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/cbethell/make-conda-default/README.md#snakemakeconda-installation). 
Package dependencies for the analysis workflows in this repository are managed using [renv](https://rstudio.github.io/renv/index.html), which can be installed independently if desired. 
Still, we recommend using Snakemake's conda integration to set up the R environment and all dependencies that the workflow will use.

```

allyhawkins (approved):
Once the few things I noted here are fixed, then we should be good to go. I don't need to see it again.

PR #366, opened 2023-08-02
Add batch ASW plots to integration report

### LINKED ISSUE #355: Add batch ASW visualization to integration report

We will want to include visualizations of the batch average silhouette width (ASW) for the unintegrated and integrated data in the html report.

A subsection under the "integration performance" section should be added specifically for "batch ASW".

A "batch ASW" section should be added to the `integration-report-template.Rmd` with the following:

- steps to source or run a utils function specific to calculating batch ASW
- a plot with the batch ASW values on the y-axis and the datasets on the x-axis (unintegrated, fastMNN, and/or harmony)

cbethell:
**Issue Addressed**
Closes #355

The purpose of this PR is to add function implementation to calculate silhouette width to the integration report, and to add the plots to display the calculate values.

- A function was added to `utils/integration-functions.R` named `plot_asw()` -- this function summarizes/averages the silhouette width values and plots them
- A `Batch ASW plots` section was added to the integration report template which implements the `calculate_silhouette_width()` function from `scpcaTools` before implementing the `plot_asw()` function mentioned above

This works for me locally as I have the development version of `scpcaTools` but upon my attempt to knit the HTML, I got the following error:
Screen Shot 2023-08-04 at 9 36 43 AM

- Does this template report knit for you with the above mentioned error? 
- Do the plots looks reasonable?

**Checklist**


allyhawkins:
> This works for me locally as I have the development version of `scpcaTools` but upon my attempt to knit the HTML, I got the following error:
> Screen Shot 2023-08-04 at 9 36 43 AM

I think you will need to update the version of `scpcaTools` that's in the lock file. Right now it's set at `v0.2.1` which won't have the most recent changes. We might want to tag a new release with these additions so that we can update to use that new version.

optional-integration-analysis/integration-report-template.Rmd
```diff
@@ -130,6 +131,41 @@ ggpubr::ggarrange(plotlist = batch_umaps,
...
+Therefore, a well-integrated dataset would show lower batch ASW values compared to an unintegrated or poorly-integrated dataset.
+
+```{r}
+# Prepare vector of PC names
+pc_names <-  stringr::str_split(params$integration_method, ",") %>%
+  unlist() |>
+  stringr::str_trim() |>
+  purrr::map(\(method) paste(method, "PCA", sep = "_"))
+
+# Set PC name for the unintegrated
```
allyhawkins, line 150:
```suggestion
# Add unintegrated PCA to pc names
```
Just clarifying what's being done here in the comment.

utils/integration-functions.R
```diff
@@ -107,3 +107,109 @@ add_integrated_pcs <- function(merged_sce,
   return(merged_sce)
   
 }
+
+plot_asw <- function(asw_df,
+                     seed = NULL, 
+                     by_batch,
```
allyhawkins, line 113: I think we can get rid of this and always color by batch.

utils/integration-functions.R
```diff
@@ -107,3 +107,109 @@ add_integrated_pcs <- function(merged_sce,
...
+        fun.min = function(x) {
+          quantile(x, 0.25)
+        },
+        fun.max = function(x) {
+          quantile(x, 0.75)
+        },
+        geom = "pointrange",
+        size = 0.3
+      )
+  }
```
allyhawkins, line 203: So I think you can get rid of all of this and always just plot with labeling the batches.

utils/integration-functions.R
```diff
@@ -107,3 +107,109 @@ add_integrated_pcs <- function(merged_sce,
...
+        size = 0.2
+      ) 
+    
+    if (is.null(plot_colors)) {
+      asw_plot <- asw_plot + 
+        ggokabeito::scale_color_okabe_ito(name = batch_label) 
+    } else {
+      asw_plot <- asw_plot + 
+        scale_color_manual(name = batch_label, values = plot_colors, labels = legend_labels)  
+    }
```
allyhawkins, line 175: We had only been using this in scenarios where we had a large amount of cell types and were coloring by cell type. I don't think we need this here and can remove the `plot_colors` and `legend_labels` arguments. 

This may be a problem when we have a lot of batches, but I think for now we can leave it.

utils/integration-functions.R
```diff
@@ -107,3 +107,109 @@ add_integrated_pcs <- function(merged_sce,
   return(merged_sce)
   
 }
+
+plot_asw <- function(asw_df,
+                     seed = NULL, 
+                     by_batch, 
+                     batch_label, 
+                     plot_colors = NULL, 
+                     legend_labels = NULL) {
```
allyhawkins, line 116:
```suggestion
                     batch_label) {
```
Based on my comment below, remove these.

allyhawkins (commented):
Just a few comments to help simplify this function. Also can you include an example of the updated plot when you address these comments? That way it will be easier to review and make sure it looks as expected.

allyhawkins:
We specify the version of `scpcaTools` to use when setting up the conda environments for the workflows in this repo. 
https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/cbcea03855d89a93deb7d14621fecf85a06e9250/envs/scpca-renv.post-deploy.sh#L9

What I think we will need to do then is first release a new version of `scpcaTools` and then update the version used here. @jashapiro are you okay with a new release of `scpcaTools` or do you have any other ideas?

optional-integration-analysis/integration-report-template.Rmd
```diff
@@ -130,6 +131,41 @@ ggpubr::ggarrange(plotlist = batch_umaps,
...
+Batch silhouette width is calculated across 20 replicates of downsampling, and the mean silhouette widths are calculated (ASW) and shown below.
+The batch silhouette width therefore measures to what extent a given data point adheres to its batch group, which is not desired after integration.
+Therefore, a well-integrated dataset would show lower batch ASW values compared to an unintegrated or poorly-integrated dataset.
+
+```{r}
+# Prepare vector of PC names
+pc_names <-  stringr::str_split(params$integration_method, ",") %>%
+  unlist() |>
+  stringr::str_trim() |>
+  purrr::map(\(method) paste(method, "PCA", sep = "_"))
```
jashapiro, line 148: paste is already vectorized, so we don't need `map` here, and we can simplify a bit with `paste0()` since we are just adding a suffix,
```suggestion
  paste0("_PCA",)
```

jashapiro:
> > This works for me locally as I have the development version of `scpcaTools` but upon my attempt to knit the HTML, I got the following error:
> > Screen Shot 2023-08-04 at 9 36 43 AM
> 
> I think you will need to update the version of `scpcaTools` that's in the lock file. Right now it's set at `v0.2.1` which won't have the most recent changes. We might want to tag a new release with these additions so that we can update to use that new version.

Since this is going into `development` branch, I think we are fine to use the `:edge` tag.

optional-integration-analysis/integration-report-template.Rmd
```diff
@@ -63,6 +63,7 @@ setup_renv(project_filepath = project_root)
 library(SingleCellExperiment)
 library(ggpubr)
 library(dplyr)
+library(scpcaTools)
```
jashapiro, line 66: Do we want to import the whole package or use `scpcaTools::` throughout. I would tend to vote for the latter for clarity.

utils/integration-functions.R
```diff
@@ -107,3 +107,109 @@ add_integrated_pcs <- function(merged_sce,
...
+  
+  # Check that all expected columns are present in dataframe
+  expected_columns <- c("pc_name", "rep", "silhouette_width", "silhouette_cluster")
+  if(!all(expected_columns%in% colnames(asw_df))){
+    stop("Required columns are missing from input dataframe, make sure that `calculate_silhouette_width` has been run successfully.")
+  }
+  
+  # Prepare and plot data by batch
+  if (by_batch) {
+    asw_plot <- asw_df %>%
```
jashapiro, line 139: for style: are we using `%>%` or `|>` in this repo? Noting that `%>%` does not seem to be imported in this file, so you should probably stick with `|>`? 

Separately though, `library(ggplot2)` should be added here for the plotting functions that you are using.

utils/integration-functions.R
jashapiro: I agree with this!

jashapiro (commented):
I agree with Ally's comments here, and added a few more little things.

jashapiro:
> We specify the version of `scpcaTools` to use when setting up the conda environments for the workflows in this repo.
> 
> https://github.com/AlexsLemonade/scpca-downstream-analyses/blob/cbcea03855d89a93deb7d14621fecf85a06e9250/envs/scpca-renv.post-deploy.sh#L9
> 
> What I think we will need to do then is first release a new version of `scpcaTools` and then update the version used here. @jashapiro are you okay with a new release of `scpcaTools` or do you have any other ideas?

Oh, right, I was still in docker land. I think you can just change that to point at `"main"` for the development branch. No need to wait for a new release.

allyhawkins:
> Oh, right, I was still in docker land. I think you can just change that to point at "main" for the development branch. No need to wait for a new release.

So change `SCPCATOOLS_VERS='v0.2.1'` to `SCPCATOOLS_VERS='main'`?

jashapiro:
> > Oh, right, I was still in docker land. I think you can just change that to point at "main" for the development branch. No need to wait for a new release.
> 
> So change `SCPCATOOLS_VERS='v0.2.1'` to `SCPCATOOLS_VERS='main'`?

I would expect that to work, yes. We would also want to make sure to update the `renv.lock` afterward with a call to the `setup_envs.sh` script.

utils/integration-functions.R
```diff
@@ -107,3 +107,66 @@ add_integrated_pcs <- function(merged_sce,
...
+      fun.min = function(x) {
+        quantile(x, 0.25)
+      },
+      fun.max = function(x) {
+        quantile(x, 0.75)
+      },
+      geom = "pointrange",
+      position = position_dodge(width = 0.5),
+      size = 0.2
+    )
```
allyhawkins, line 160: Can we add a theme here and do `theme_bw()`? The UMAPS are like that so if it's not set at the beginning of the notebook, we should do that.

utils/integration-functions.R
```diff
@@ -107,3 +107,66 @@ add_integrated_pcs <- function(merged_sce,
   return(merged_sce)
   
 }
+
+plot_asw <- function(asw_df,
```
allyhawkins, line 111: in this function you are using `%>%` and in the integration report you are using `|>`. I think we should pick one and be consistent. Maybe for the integration module we can use `|>` since we have already been using that here.

allyhawkins (commented):
Thanks for attaching the report! I had a few suggestions to improve the plot.

integration.snakefile
```diff
@@ -56,3 +59,27 @@ rule perform_integration:
...
+        integrated_sce = "{basedir}/{group}_integrated_sce.rds"
+    output:
+        "{basedir}/{group}_integration_report.html"
+    log: "logs/{basedir}/{group}/integration_report.log"
+    conda: "envs/scpca-renv.yaml"
+    shell:
+        """
+        Rscript --vanilla -e "
+          source(file.path('$PWD', 'utils', 'setup-functions.R'))
+          setup_renv(project_filepath = '$PWD')
```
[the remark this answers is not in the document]
cbethell: Good point, removed this in https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/366/commits/ca24f2c6860a32e6a8d4fda95974561fdb6662ba

utils/integration-functions.R
cbethell: I set the theme at the beginning of the notebook but also added `theme_bw()` to the UMAP plotting function as it seems that the UMAP plotting was over-riding the theme set.

cbethell:
> Thanks for attaching the report! I had a few suggestions to improve the plot.

Here is the updated report with those changes @allyhawkins 👍 
[group01_integration_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/12294965/group01_integration_report.html.zip)

jashapiro (approved):
LGTM

PR #368, opened 2023-08-07
Add within-batch ARI plots to report template

### LINKED ISSUE #354: Add within-batch ARI visualization to integration report

We will want to include visualizations of the within-batch ARI performance metric output on the unintegrated and integrated data in the html report.

A subsection under the "integration performance" section should be added specifically for "within-batch ARI".

A "within-batch ARI" section should be added to the `integration-report-template.Rmd` with the following:

- steps to source or run a utils function specific to calculating the within-batch ARI
- a plot showing the within-batch ARI values on the y-axis and the dataset on the x-axis (unintegrated, fastMNN, and/or harmony)

We can add the function for calculating the within-batch ARI values to the ARI-specific script in utils like in the `sc-data-integration` repo: https://github.com/AlexsLemonade/sc-data-integration/blob/main/scripts/utils/calculate-ARI.R

cbethell:
**Issue Addressed**
Closes #354

The purpose of this PR is to add function implementation to calculate within-batch ARI values to the integration report, and to add the plots to display the calculated values.

A within-batch ARI plots section was added to the integration report template which implements the calculate_within_batch_ari() function from `scpcaTools` 

**Note**: When trying to knit/run the template locally, I am receiving an error that object `batch_ids` cannot be found -- I tried adjusting the `RemoteSha` hash in the lock file to the latest commit for scpcaTools (which fixes the batch ids issue) but still receive the same error. I also tried running the `setup_envs.sh` script once again but that did not work. Any suggestions @allyhawkins @jashapiro?

**Checklist**

### DIFF
--- a/optional-integration-analysis/integration-report-template.Rmd
+++ b/optional-integration-analysis/integration-report-template.Rmd
@@ -177,7 +177,9 @@ Within-batch ARI values close to 1 indicate that biological heterogeneity has be
 ```{r}
 # List of SCE filepaths
 input_metadata <- readr::read_tsv(params$input_metadata_tsv)
-sce_files <- input_metadata$processed_sce_filepath
+sce_files <- input_metadata |>
+  dplyr::filter(integration_group == params$integration_group) |>
+  dplyr::pull(processed_sce_filepath)
 # Read in list of SCEs
 sce_list <- purrr::map(sce_files, readr::read_rds) |>
   purrr::set_names(input_metadata$library_id)
@@ -191,14 +193,23 @@ within_batch_ari_df <-
     batch_column = params$batch_column,
     BLUSPARAM = bluster::NNGraphParam(cluster.fun = "louvain", type = "jaccard"),
     seed = params$seed
-  )
-
-within_batch_ari_df$ari <- as.numeric(within_batch_ari_df$ari)
+ ) |>
+  dplyr::mutate(ari = as.numeric(ari))
                                                   
 ```
 
 ```{r}
-ggplot(within_batch_ari_df, aes(x = pc_name , y = ari, color = batch_id)) +
+within_batch_ari_plot <- within_batch_ari_df |>
+  dplyr::mutate(
+    integration_method = dplyr::if_else(
+      pc_name == "PCA",
+      "Pre-Integration",
+      stringr::str_remove(pc_name, "_PCA")
+    ),
+    integration_method_factor = forcats::fct_relevel(integration_method, "Pre-Integration")
+  ) |>
+  ggplot() +
+  aes(x = pc_name , y = ari, color = batch_id) +
   geom_point() +
   ggforce::geom_sina() +
   # add median point to plot
@@ -214,9 +225,15 @@ ggplot(within_batch_ari_df, aes(x = pc_name , y = ari, color = batch_id)) +
     geom = "pointrange",
     size = 0.4
   ) +
+  guides(color = guide_legend(
+    override.aes = list(size = 3),
+    label.theme = element_text(size = 12)
+  )) +
   labs(x = "Integration method",
        y = "ARI",
        color = "Batch")
+
+within_batch_ari_plot
 ```
 
 ## Session info
--- a/optional-integration-analysis/integration-report-template.Rmd
+++ b/optional-integration-analysis/integration-report-template.Rmd
@@ -176,9 +176,9 @@ Within-batch ARI values close to 1 indicate that biological heterogeneity has be
 
 ```{r}
 # List of SCE filepaths
-input_metadata <- readr::read_tsv(params$input_metadata_tsv)
+input_metadata <- readr::read_tsv(params$input_metadata_tsv) |> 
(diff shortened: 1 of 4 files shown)


optional-integration-analysis/integration-report-template.Rmd
```diff
@@ -164,6 +165,60 @@ plot_asw(asw_df = asw_results,
...
+The metric measures how well a given integration method preserves biological heterogeneity by comparing pre- and post-integration clustering assignments for each library.
+We take each original SCE object, prior to any merging, and perform graph-based clustering.
+The same clustering is then performed on the integrated object.
+For each library that is present in the integrated object, the original cluster assignments are compared to the cells’ new cluster assignments post integration by calculating the ARI between these two cluster assignments.
+Within-batch ARI values close to 1 indicate that biological heterogeneity has been preserved.
+
+```{r}
+# List of SCE filepaths
+input_metadata <- readr::read_tsv(params$input_metadata_tsv)
+sce_files <- input_metadata$processed_sce_filepath
```
allyhawkins, line 180: I think you'll want to filter these to only include the sce objects that are in the specified group.

allyhawkins (commented):
I'm leaving some initial thoughts here. As far as the error, my only thought right now is deleting the conda folder like I mentioned in slack.

cbethell:
> I'm leaving some initial thoughts here. As far as the error, my only thought right now is deleting the conda folder like I mentioned in slack.

Thanks @allyhawkins, deleting the conda folder worked 🎉 I was able to implement the comments you left above as well.

Here's the example rendered report for reference:
[group01_integration_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/12304232/group01_integration_report.html.zip)

optional-integration-analysis/integration-report-template.Rmd
```diff
@@ -164,6 +165,77 @@ plot_asw(asw_df = asw_results,
...
+
+```{r}
+# List of SCE filepaths
+input_metadata <- readr::read_tsv(params$input_metadata_tsv)
+sce_files <- input_metadata |>
+  dplyr::filter(integration_group == params$integration_group) |>
+  dplyr::pull(processed_sce_filepath)
+# Read in list of SCEs
+sce_list <- purrr::map(sce_files, readr::read_rds) |>
+  purrr::set_names(input_metadata$library_id)
```
allyhawkins, line 185:
```suggestion
input_metadata <- readr::read_tsv(params$input_metadata_tsv) |> 
  dplyr::filter(integration_group == params$integration_group)
sce_files <- input_metadata |>
  dplyr::pull(processed_sce_filepath)
# Read in list of SCEs
sce_list <- purrr::map(sce_files, readr::read_rds) |>
  purrr::set_names(input_metadata$library_id)
```
I think I would slightly switch this so that you filter first and then you can use that to both define the sce files and the library ids.

allyhawkins (commented):
A few more small comments to address. The other thing I'm a little concerned about is that there looks to be two dots for each library with the exact same ari value? I think there should only be 1 value for each batch for each integration method correct?

cbethell:
> A few more small comments to address. The other thing I'm a little concerned about is that there looks to be two dots for each library with the exact same ari value? I think there should only be 1 value for each batch for each integration method correct?

Fixed the two dots issue by removing the additional `geom_point()` layer and also accounted for the `cell_id_column` argument that was added to the within-batch scpcaTools function 👍 

For reference: 
[group01_integration_report.html.zip](https://github.com/AlexsLemonade/scpca-downstream-analyses/files/12305541/group01_integration_report.html.zip)

optional-integration-analysis/integration-report-template.Rmd
```diff
@@ -164,6 +166,77 @@ plot_asw(asw_df = asw_results,
...
+      stringr::str_remove(pc_name, "_PCA")
+    ),
+    integration_method_factor = forcats::fct_relevel(integration_method, "Pre-Integration")
+  ) |>
+  ggplot() +
+  aes(x = integration_method_factor, y = ari, color = batch_id) +
+  ggforce::geom_sina() +
+  # add median point to plot
+  stat_summary(
+    color = "red",
```
allyhawkins, line 218:
```suggestion
    color = "black",
```
I think we should match the aesthetics between the two plots as much as possible.

optional-integration-analysis/integration-report-template.Rmd
```diff
@@ -164,6 +166,77 @@ plot_asw(asw_df = asw_results,
...
+    fun = "median",
+    fun.min = function(x) {
+      quantile(x, 0.25)
+    },
+    fun.max = function(x) {
+      quantile(x, 0.75)
+    },
+    geom = "pointrange",
+    size = 0.4
```
allyhawkins, line 227:
```suggestion
    size = 0.2
```

allyhawkins (approved):
Looks good, my only remaining comment is just to change the color of the median. But I don't need to see it again.

PR #371, opened 2023-08-11
Fix the conda default

cbethell:

Upon testing the integration workflow in preparation for release, I realized `use-conda` is not actually a default just yet, this PR ensures that is a default.

The following was added at the top of each snakefile and successfully tested:

```
if 'use-conda' in config and config['use-conda']:
    workflow.use_conda = True
```

**Checklist**


allyhawkins (approved):
🚀

PR #376, opened 2023-08-14
Jashapiro/env update

### LINKED ISSUE #375: Update snakemake version docs
Today I realized that neither `conda` nor `snakemake` were installed in my server account, so I got myself some fresh installs. Then I went to build the environment for this repo to work on some things, but got a surprising error:

Screenshot 2023-08-14 at 1 02 11 PM

From my fresh install, I ended up with `python 3.11` and `snakemake 7.32.3`, both of which satisfy our current docs requirements, but clearly something is up! So..

- First, I downgraded python to `3.9`. The error persisted
- Then, I downgraded `snakemake` to `7.24.2`, which is compatible with our docs (>7.20.0), and also that's the version @allyhawkins has so it seemed like a good place for success. Indeed **the error went away** and I built the environment successfully.

So, seems something happened _probably_ in the `7.3` release of snakemake (I don't know that for sure, but based on semantic versioning, probably?). 
The quickest fix here is simply to update our docs that snakemake needs to be `\in [7.20.0, 7.3)`, but more generally we might want to think about workflow compatibility (as a separate much larger issue, which depends also on the future direction of this project!).

jashapiro:
**Issue Addressed**
Closes #375

Setting `workflow.use_conda` seemed to cause trouble depending on the version of snakemake (See #375), and there were potentially some other problems, so I'm trying to get things back to working.

The main change is reverting #371 and #365.
I also updated readmes to reflect current versions in the conda/renv files.
Finally, I took the chance to update the `renv` to 1.0.0, just for maybe some future reproducibility, which means that a few new packages were required. Look at the names of those for some little treats.

I had tried to look at just fixing #375 without reversions, but I can't find any documentation on the "right" way to set `use-conda` by default, so I think it is better to just require it to be done manually (or in an wrapper script)

While this "works for me" in so far as the conda setup steps do run as expected both locally and on the server, I don't seem to be a robust test on my own. Everybody should download and try to install this locally using `conda`.

If we want to preserve the ability to run with renv, I suspect there may still be some `renv`-specific challenges that remain related to arm processors and older packages that might need to be compiled from source (I ran into a problem with `RCurl` on my machine, for example, but didn't try to fix it here). Fixing those kinds of issues probably would require updating the package versions in the `conda` environment to ones that we know work on M1 macs (and hopefully can be installed as binaries).

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@
 - [Additional analysis modules](#additional-analysis-modules)
   - [Clustering analysis](#clustering-analysis)
   - [Genes of interest analysis](#genes-of-interest-analysis)
+  - [Data integration analysis](#data-integration-analysis)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -45,8 +46,8 @@ Cluster assignments are stored in the `SingleCellExperiment` object returned by 
 
 You can read more details about the individual steps of the workflow in the processing documentation linked below:
 
-|[View Processing Information Documentation](./additional-docs/processing-information.md)|
-|---|
+| [View Processing Information Documentation](./additional-docs/processing-information.md) |
+| ---------------------------------------------------------------------------------------- |
 
 ## Quick Start Guide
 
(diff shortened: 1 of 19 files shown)


sjspielman (approved):
LGTM! Thanks for fixing this!!

Noting this review based on success running in a stacked branch - https://github.com/AlexsLemonade/scpca-downstream-analyses/pull/377#pullrequestreview-1578538972

allyhawkins (approved):
Approving based on testing #377

PR #377, opened 2023-08-14
374 Update packages to fix error when building conda environment

### LINKED ISSUE #374: Update packages to fix error when building conda environment
In testing that the workflow was good to go for release, Chante noticed an error when building the conda environment. Specifically, the version we were using of `GenomeInfoDbData` is outdated and the package download is failing. One solution is to update the version that we use to the most [recent version of `GenomeInfoDBData`](https://bioconda.github.io/recipes/bioconductor-genomeinfodbdata/README.html), but that relies on R 4.3 so we would also need to update base R (and other packages presumably). 

My messages with Chante are gone sadly, but Stephanie was looking into this and also found the same error. I'm including that here for reference. 
image

jashapiro:
stacked on #376

closes #374

Here I cleaned up package installation to work when the CRAN mirror is not set by adding a `repos` argument to `remotes::install_version()`, which was causing at least some of the errors seen in #374.

I also removed the "on-the-fly" installation of `gt`, as we can install that through conda/renv just fine, and that installation was already part of #376.

Finally, I moved some package loading calls to after the `setup_renv` calls to prevent potential errors with incorrect versions of packages being loaded (or not  set if `renv` installation was used.

Same caveats apply here as in #376: This all seems to work for me, but I am not representative of all users!

I tested this by running 
```
snakemake --use-conda -s integration.snakefile -c4
```
Which ran without incident both locally and on the server.

### DIFF
--- a/components/dependencies.R
+++ b/components/dependencies.R
@@ -7,5 +7,6 @@
 #
 # library(dplyr)
 #
-library(uwot)
+library(remotes)
 library(markdown)
+library(uwot)
--- a/envs/scpca-renv.post-deploy.sh
+++ b/envs/scpca-renv.post-deploy.sh
@@ -8,9 +8,18 @@ fi
 
 SCPCATOOLS_VERS='main'
 
-# install github packages
+# install packages not on conda-forge
 Rscript --vanilla -e \
   "
+  # install Harmony from CRAN
+  remotes::install_version(
+    'harmony', 
+    version = '0.1.1', 
+    repos = 'https://cloud.r-project.org', 
+    upgrade = FALSE
+  )
+
+  # install ScPCA tools from Github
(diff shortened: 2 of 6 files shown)


jashapiro:
I also moved `harmony` installation to within the environment setup, which seemed to be causing me trouble, until I remembered something long past about a conda package, `conda-ecosystem-user-package-isolation`, that isolates user libraries: I have now added that to the conda repo, and it seems to have solved the issues I was seeing. Maybe it will help others too and make things more reliable! 🤞🏼

sjspielman (approved):
From _this_ branch `jashapiro/374-fix-package-install`, I ran 
```
bash setup-envs.sh clean
```

both locally and on the server. Immediate success all around 🎉

sjspielman:
Also wanted to say A+ for `conda-ecosystem-user-package-isolation`!

allyhawkins (approved):
Success 🎉

